PR #113, opened 2015-01-07
Add 'alt' attribute on all images that were missing

marcker:
With 'alt' attributes this now is a [valid HTML5]("http://validator.w3.org/") website.


jina:
I'm usually strict about adding those, but in the time crunch of launching the site, seems I totally skipped those. Oops!!! :) Thanks for catching that!

marcker:
It was a pleasure to contribute.

PR #117, opened 2015-02-02
resizing bg prism to get rid of weird gap on retina

GarthDB:
Getting an odd gap on retina mbp from the resizing happening on the svg. Here's a fix.

before:
image


jina:
Thanks!

PR #119, opened 2015-02-04
Add Omniref instant documentation navigation.

montanalow:
Omniref is using scss liberally, and we'd love to find ways to contribute to the project. I've created a dynamic documentation menu to kick things off, but I think there's a lot more we can do to help out. In the meantime, I've poured myself some Old Weller 107, and poked around to make sure there was no perceivable change until hovering over the Documentation link. The dynamic navigation menu supports the existing responsive layout with the native sass styles and will be automatically updated whenever a new version of sass is published via ruby gems.

I've setup an idle dyno app on heroku if you'd like to play with the functionality (first request will be slow):

https://sass-omniref.herokuapp.com/

I'd appreciate any feedback you have.

Thanks!

### DIFF
--- a/source/layouts/head/_js.haml
+++ b/source/layouts/head/_js.haml
@@ -1,5 +1,7 @@
 = javascript_include_tag "vendor/modernizr-2.8.3.custom"
 = javascript_include_tag  "//use.typekit.net/pya1hgl.js"
+= javascript_include_tag "https://www.omniref.com/nav.min.js"
+
 :javascript
   try{Typekit.load();}catch(e){}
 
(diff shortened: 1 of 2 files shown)


jina:
Hi Montana! Thanks for going through this effort and for following the "drink whisky" step. ;) However, we want to keep our documentation on our own site, and as the lead designer on the website, I'm not really into the idea of introducing drop down menus. It doesn't fit within the design aesthetic, nor the simplicity we're aiming for interaction-wise. Thanks anyway! :)

montanalow:
Thanks for getting back to me so quickly. I'm a programmer not a designer,
so I'd love any thoughts you have to make navigating docs simpler. We're
looking to be more interactive by tying in discussion about the code along
side the docs (Q&A, open code review). So often the answers I need are
actually in the source code, unfortunately not documented. We've had a
little bit of success with the larger rails and ruby standard library
communities, but I'd love to know if there's anything we could build that
you'd find valuable. To keep the docs on sass-lang.com, would you consider
setting a dns record like docs.sass-lang.com refer to Omniref docs / Q&A?

jina:
Thanks Montana. This isn't something we feel our project needs right now. We're pretty happy with our current set up. :)

PR #120, opened 2015-02-06
Updating glasses.svg styling 

heitzke:
Not a departure from the current style, just felt like it could stand to have a couple points knocked off, overlaps adjusted and some lines smoothed out.  

Unfortunately I can't seem to get this thing compressed anymore and to the current live weight, so it's ~.9k heavier from the extra points I added. I could probably simplify it some more moving forward.


heitzke:
Naturally, images help: 

before-after

jina:
I appreciate that you went through this effort. But, I do not wish to get rid of the semi colon on the right detail? It was intentional. :)



—

heitzke:
Ah! No prob, I considered that but thought it was an artifact. I'll maintain that detail and update the pr

heitzke:
Figured that i'd touch on this since it's floating out in the ether. I can go ahead and close this if we're not wanting to merge it in.

PR #126, opened 2015-03-10
Some changes

mikaspell:
bundle update, as the eventmachine version 1.0.3 is not installing

fixed sitemap bug


jina:
I think you have to update the Gemfile, not the Gemfile.lock, correct?

mikaspell:
is not correct, it is a change in gemfile.lock enables successfully run bundle install

jina:
but gemfile is the working file. gemfile.lock is generated from gemfile. gemfile.lock is not supposed to be manually edited as far as I know?


—

mikaspell:
yes you are right, if you run 'bundle update' then gemfile.lock will be overwritten automatically , and it just keeps the necessary versions of gems. But when running the command 'bundle install' is being held gemfile.lock and installs the required version of the gems. Thus, I want to say that you can manually change the file if you know exactly what version of gem you need. Let me make another pull request after executing the command 'bundle update' as there are not only upgraded eventmachine

mikaspell:
updated pull

PR #135, opened 2015-08-04
Update libSass page

oesgalha:
According to the ruby-libsass README (https://github.com/sass/ruby-libsass) that port of libSass to ruby is deprecated and sassc-ruby (https://github.com/sass/sassc-ruby) should be used instead.
This PR updates the libSass page to link to sassc-ruby as the ruby port instead of ruby-libssas.


jina:
thanks!

PR #136, opened 2015-08-06
code-snippets, fix, comment type

Mr21:
Hi :)

Here: http://sass-lang.com/guide in the *Import* section, we can see a specific change about a comment between the `Sass` and the `SCSS` version of the code.

I suppose it's a mistake, because it's give the impression to be another difference between `Sass/SCSS` but not.


jina:
thx!

PR #143, opened 2015-10-14
Update CONTRIBUTING.md

arhoads:
Changed instructions to use the bundled middleman. It is likely the gem isn't globally installed and it isn't at least if just these instructions are followed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ navigate to the project repo and run:
 
 ```
 bundle install
-middleman
+bundle exec middleman
 ```
 
 


jina:
This still needed? I updated the gem today and didn't have to preface with `bundle exec` after doing so. Can you try again and confirm?

arhoads:
I only have it installed through bundler and the Gemfile specifies ``middleman 3.4.0``. If I install it globally with ``gem install middleman`` it works without ``bundle exec``, but as the instructions are it doesn't run for me without ``bundle exec``.

jina:
okie dokie :)

Issue #132, opened 2015-05-29
Apostrophes are showing as ??? in browsers on Yosemite


headquarters:
I'm running Yosemite with Chrome 43 and noticing that apostrophes are showing as 3 question marks ("???") on all the documentation pages, such as: 
http://sass-lang.com/documentation/file.SASS_REFERENCE.html#mixins

I don't mind fixing this and providing a pull request, but I can't find the actual doc files in the repo. I've ran bundle install and middleman, but it looks like the files are pulled from the URL http://sass-lang.com/documentation/file.SASS_REFERENCE.html. Is there another repo with the doc files?

hariganesan:
Appears seems to show up on all operating systems/browsers

jina:
must be an encoding issue. @nex3 is this something we fix on the doc repo side of things?

nex3:
There might be a way to force it to always use UTF-8, but I'm not entirely sure what that would be.

adamzerner:
I'm having the same problem with the most recent version of Chrome and OSX.

nex3:
I'm pretty sure this is caused by some aspect of the configuration on certain computers that are used to deploy. When I deploy, it doesn't seem to happen, but when other people do it does. I'm still not sure how to fix it, though.

arhoads:
I am still getting this problem across browsers and OSes.

jina:
@nex3 Is there any chance we can replace `<meta http-equiv=Content-Type content="text/html; charset=US-ASCII"/>` with `<meta charset="utf-8" />` in the docs template? I cloned the [sass repository](https://github.com/sass/sass) and I searched in files, but I can not for the life of me find where the template's markup is coming from to make the switch. This would fix the issue.

nex3:
@jina You might have to hack the default YARD templates. It's been years since I did that last and I don't remember how.

headquarters:
This appears to be fixed in the most recent deployment, so I'm closing this issue for now.

PR #144, opened 2015-10-14
Specify Ruby version to avoid encoding errors

arhoads:
#132 is caused by using an old Ruby when generating the docs. I have tested with Ruby 1.9 versus Ruby 2. The error shows up in 1.9, but not 2. 

Ruby starting with 2.0 does UTF-8 encoding by default. Earlier Ruby versions do not and this can cause the unicode quotes to turn into "???" when the docs are generated with an older Ruby.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ following:
 This site is built with [middleman](http://middlemanapp.com), a Ruby framework
 for building static sites.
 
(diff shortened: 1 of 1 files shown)


jina:
@nex3 can you check this one? :)

jina:
(Thanks for looking into it, @arhoads!)

nex3:
Sounds good to me. This won't enforce Ruby 2 being used, but it may help remind us.

PR #145, opened 2015-10-16
Enforce at least Ruby 2 in rake task

arhoads:
Is this what you were looking for? @nex3


nex3:
I wasn't thinking of anything in particular, but this does seem like a good idea.

PR #148, opened 2015-10-27
fix(libsass.html): correcting link to sass.js playground

rodneyrehm:
… just fixing the link target for "testing sass.js in the browser" :)

### DIFF
--- a/source/libsass.html.haml
+++ b/source/libsass.html.haml
@@ -51,7 +51,7 @@ title: libSass
   %li#javascript
     :markdown
       ### JavaScript
-      The only pure JavaScript implementation is [sass.js](https://github.com/medialize/sass.js). There's a [browser-based way to test it out](http://medialize.github.io/sass.js/), too.
+      The only pure JavaScript implementation is [sass.js](https://github.com/medialize/sass.js). There's a [browser-based way to test it out](http://medialize.github.io/playground.sass.js/), too.
 
   %li#lua
     :markdown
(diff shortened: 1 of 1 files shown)


source/libsass.html.haml
```diff
@@ -51,7 +51,7 @@ title: libSass
   %li#javascript
     :markdown
       ### JavaScript
-      The only pure JavaScript implementation is [sass.js](https://github.com/medialize/sass.js). There's a [browser-based way to test it out](http://medialize.github.io/sass.js/), too.
+      The only pure JavaScript implementation is [sass.js](https://github.com/medialize/sass.js). There's a [browser-based way to test it out](http://medialize.github.io/playground.sass.js/), too.
```
jina: Hey @rodneyrehm  — do you mind fixing the houndci error in this and then I'll merge it in? Just make sure the code is around 78-ish characters.

rodneyrehm: I'm not familiar with haml. will a simple indented line-break do the job?

jina: Yep. It's actually in a markdown filter. Just a line break will do. Make sure it's at the same level of indention though. Thanks!

rodneyrehm:
@jina there you go, 80 characters. Note that pretty much every other line in that file is *way* longer than that limit. Are you sure imposing limits like that on flowing text content is a good idea?

jina:
It's been in the [Code Style Guide](http://sass-lang.com/styleguide/code) from the launch of the site, but no one has been paying attention to it.

The HoundCI checker will help make sure new pull requests going forward follow the guidelines.

It's for better code quality. :)

PR #156, opened 2015-12-30
Reduce lagging in Safari 9

leo:
When trying to resize the window containing the site, the browser will suddenly start lagging behind. The reason for this is the SVG background being repeated endlessly.

This PR heavily reduces the lagging problem by only repeating the background on the x-axis for big screens.

### DIFF
--- a/source/assets/css/foundation/_site.scss
+++ b/source/assets/css/foundation/_site.scss
@@ -1,10 +1,20 @@
-* { @include box-sizing(border-box); }
+* {
+  @include box-sizing(border-box);
+}
 
 @include establish-baseline;
 
-%reset-margin  { margin: 0; }
-%reset-border  { border: 0; }
-%reset-padding { padding: 0; }
+%reset-margin {
+  margin: 0;
+}
+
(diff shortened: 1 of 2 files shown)


jina:
Thanks for this (and for fixing the HoundCI issues, too). :)

source/assets/css/foundation/_site.scss
```diff
@@ -14,26 +24,47 @@
...
+  @include theme-selection;
+}
 
+::selection {
+  @include theme-selection;
+}
 
 @include breakpoint($tablet-large) {
+
   html,
```
jina: I went ahead and merged your PR, but just noticed that you changed the selectors from being on their own line — prefer them to stay as is. Will change it back later. :)

leo: What do you mean? In my first commit on this PR, I've changed them both to be in one single line. But HoundCI told me to put them in separate lines, so... :blush:

jina: oh I misread the history. :)

PR #157, opened 2016-01-20
rewrote GoLang links

tomByrer:
removed abandoned Go Libsass wrappers
added pure GoLang compilers

### DIFF
--- a/source/libsass.html.haml
+++ b/source/libsass.html.haml
@@ -37,11 +37,17 @@ title: libSass
   %li#go
     :markdown
       ### Go
-      [Wellington](https://github.com/wellington/wellington) is an extension to libsass that adds spriting and is available on brew:
+      [go-libsass](https://github.com/wellington/go-libsass) has the most active
+      GoLang wrapper.  [gosass](https://github.com/moovweb/gosass) is another
+      libsass wrapper.
+      
+      [Wellington](http://getwt.io/) is an extension to libsass that adds
+      spriting.  It is available on brew: `brew install wellington`
 
-      `brew install wellington`
-
-      There are also three other libsass wrappers in go: [gosass](https://github.com/moovweb/gosass), [go-sass](https://github.com/SamWhited/go-sass) and [go_sass](https://github.com/suapapa/go_sass) which have not been updated in a while.
+      [C6](https://github.com/c9s/c6) is a SASS 3.2 compatible implementation
+      written in pure GoLang that aims to extend SASS.
+      [wellington/sass](https://github.com/wellington/sass) is an in-progress
+      pure Go sass lexer, parser, and compiler.
 
   %li#java
     :markdown


source/libsass.html.haml
```diff
@@ -37,11 +37,17 @@ title: libSass
...
       ### Go
-      [Wellington](https://github.com/wellington/wellington) is an extension to libsass that adds spriting and is available on brew:
-
-      `brew install wellington`
-
-      There are also three other libsass wrappers in go: [gosass](https://github.com/moovweb/gosass), [go-sass](https://github.com/SamWhited/go-sass) and [go_sass](https://github.com/suapapa/go_sass) which have not been updated in a while.
+      [go-libsass](https://github.com/wellington/go-libsass) has the most active
+      GoLang wrapper.  [gosass](https://github.com/moovweb/gosass) is another
+      libsass wrapper.
+
```
jina: Hi @tomByrer! I added @houndci to help keep the code quality up. Do you mind making the small suggestion made here? Thanks for your time!

xzyfer:
@jina wellington is not abandoned. We (the LibSass team) work closely with it's author @drewwells.

xzyfer:
Apologies I read the diff incorrectly.

source/libsass.html.haml
```diff
@@ -37,11 +37,17 @@ title: libSass
...
-
-      There are also three other libsass wrappers in go: [gosass](https://github.com/moovweb/gosass), [go-sass](https://github.com/SamWhited/go-sass) and [go_sass](https://github.com/suapapa/go_sass) which have not been updated in a while.
+      [go-libsass](https://github.com/wellington/go-libsass) has the most active
+      GoLang wrapper.  [gosass](https://github.com/moovweb/gosass) is another
+      libsass wrapper.
+      
+      [Wellington](http://getwt.io/) is an extension to libsass that adds
+      spriting.  It is available on brew: `brew install wellington`
+
+      [C6](https://github.com/c9s/c6) is a SASS 3.2 compatible implementation
```
xzyfer: C6 is not related to LibSass.

source/libsass.html.haml
```diff
@@ -37,11 +37,17 @@ title: libSass
   %li#go
     :markdown
       ### Go
-      [Wellington](https://github.com/wellington/wellington) is an extension to libsass that adds spriting and is available on brew:
-
-      `brew install wellington`
-
-      There are also three other libsass wrappers in go: [gosass](https://github.com/moovweb/gosass), [go-sass](https://github.com/SamWhited/go-sass) and [go_sass](https://github.com/suapapa/go_sass) which have not been updated in a while.
+      [go-libsass](https://github.com/wellington/go-libsass) has the most active
```
xzyfer: The last commit to go-libsass is almost a year ago. There have been several major LibSass releases since then. IMHO this project looks to be abandoned.

drewwells: Go-libsass was updated this week and points to master of libsass.

drewwells: The other go bindings should be removed, they look abandoned. At least one
of them directs people to go-libsass which has the most comprehensive
coverage of the libsass api.

drewwells:
Thanks for updating these links @tomByrer!

source/libsass.html.haml
xzyfer: Sorry I was looking at the original go-libsass. 

>removed abandoned Go Libsass wrappers

This makes sense now.

xzyfer: Apologies this was pre-morning-coffee

source/libsass.html.haml
tomByrer: Let's figure out final changes then I'll likely resubmit.
BTW, too bad your bot wouldn't auto-fix simple formatting changes...

source/libsass.html.haml
tomByrer: Ah, I misread the JS entry "The only pure JavaScript implementation is sass.js" which technically libsass compiled into JS JIT via emscripten.  So kinda still libsass.

Well obviously Ruby-sass & libsass aren't the only compilers in town anymore.  They should be listed somewhere...  What's the best approach?  Rename this page to "alternative Sass compilers" & rework it?

source/libsass.html.haml
tomByrer: np
I reordered to:
libsass wrappers
utilities
non-libsass parcers

source/libsass.html.haml
xzyfer: There's some background context, but long story short c6 is not a Sass implementation, it's a Sass compatible language since they've taken some freedoms with with adding additional syntax.

drewwells: My take on c6 is that c6 doesn't aim to bring Sass to go. The author sees an opportunity to fork Sass and make a new, incompatible version of the language. The project lacks a compiler, so you can't do anything but create AST for the parts of this language that has been implemented and the main developer has discontinued work on it. Even if it was finished, there is no guarantee Sass code would compile with c6 and may never compile with it. It's interesting, but would be misleading to list it as a Sass compiler. Maybe if there was a toy project section or misc tools related to Sass section.

xzyfer: Pretty much this. Please remove it.

source/libsass.html.haml
```diff
@@ -37,11 +37,17 @@ title: libSass
...
+      GoLang wrapper.  [gosass](https://github.com/moovweb/gosass) is another
+      libsass wrapper.
+      
+      [Wellington](http://getwt.io/) is an extension to libsass that adds
```
xzyfer: `libsass` ->  `LibSass`

source/libsass.html.haml
```diff
@@ -37,11 +37,17 @@ title: libSass
...
     :markdown
       ### Go
-      [Wellington](https://github.com/wellington/wellington) is an extension to libsass that adds spriting and is available on brew:
-
-      `brew install wellington`
-
-      There are also three other libsass wrappers in go: [gosass](https://github.com/moovweb/gosass), [go-sass](https://github.com/SamWhited/go-sass) and [go_sass](https://github.com/suapapa/go_sass) which have not been updated in a while.
+      [go-libsass](https://github.com/wellington/go-libsass) has the most active
+      GoLang wrapper.  [gosass](https://github.com/moovweb/gosass) is another
+      libsass wrapper.
```
xzyfer: `libsass` ->  `LibSass`

xzyfer: My impression is that `moovweb/gosass` is dead. Probably not worth mentioning.

source/libsass.html.haml
tomByrer: > Maybe if there was a toy project section or misc tools related to Sass section. 

Seems like a better idea.  Should wellington/sass also belong in there?

drewwells: Yeah that's probably a good idea. It's likely a year away from being useful to anyone.

chriseppstein: Agree with @xzyfer and @drewwells. The c6 developer has not agreed to maintain the implementation as the language evolves and it does not pass sass-spec afaik. We don't promote such projects on our site.

jina:
@xzyfer @chriseppstein where do we stand on this PR — is it approved (assuming the conflicts get merged first)? Or were there more things to be addressed?

chriseppstein:
@jina the links now appear to point to projects that haven't been updated in a while. Still, it's probably more accurate than it would be without it once the comments are addressed. I don't think we can expect the OP to do the changes at this point.

jina:
K. I'll look at the changes when I get a chance!

PR #160, opened 2016-02-16
fix typo

Youmoo:


jina:
thanks! this will be in the next deploy

PR #161, opened 2016-02-27
[RFR] Normalised LibSass capitalisation

HugoGiraudel:
Because 4 different spellings in the same page is a bit too much. :)

### DIFF
--- a/source/libsass.html.haml
+++ b/source/libsass.html.haml
@@ -1,11 +1,11 @@
 ---
-title: libSass
+title: LibSass
 ---
 
 %p.introduction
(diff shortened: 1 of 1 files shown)


jina:
@xzyfer did y'all ever figure out which spelling you wanted?

xzyfer:
@jina we've recently normalised the repo on LibSass.

jina:
@xzyfer cool. would have been good to do that here, too. ;) Thanks @HugoGiraudel! this will get deployed once i have a chance to do so from my computer. :)

PR #163, opened 2016-03-20
Updated the year in the footer

ghoshnirmalya:


source/layouts/regions/_contentinfo.haml
```diff
@@ -4,7 +4,7 @@
       %ul
         %li.contentinfo-legal
           :markdown
-            Sass &copy; 2006&ndash;2015 [Hampton Catlin](http://www.hamptoncatlin.com/),
+            Sass &copy; 2006&ndash;2016 [Hampton Catlin](http://www.hamptoncatlin.com/),
```
ghoshnirmalya: Ignoring this as it was previously of the same length.

PR #167, opened 2016-06-26
Change gem install command in Linux

GeorgeTG:
The proposed command in the install page of the site:
`sudo su -c "gem install sass"`
neglects the fact that some gem packages come with `--user-install` in the default `/etc/gemrc`.
This command will install sass in the $HOME directory of the user root which is /root.
This makes sass available *only* to the Root user.
As expected, the following warning is given:
`WARNING:  You don't have /root/.gem/ruby/2.3.0/bin in your PATH,`
`	  gem executables will not run.`
By using the `--no-user-install` option it works correctly, as the gem is installed in $GEM_HOME instead as pointed out by `gem help install`

### DIFF
--- a/source/install.html.haml
+++ b/source/install.html.haml
@@ -66,7 +66,7 @@ title: Install Sass
         first. You can install Ruby through the apt package manager, rbenv, or rvm.
         %pre
           :preserve
-            sudo su -c "gem install sass"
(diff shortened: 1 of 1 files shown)


jina:
@nex3 or @chriseppstein — can you review this one? I’m not knowledgeable about this. :)

PR #173, opened 2016-09-06
Scout to Scout-App, OS Support updated

TheJaredWilcurt:
Scout has been rebranded to Scout-App. OS Support now includes Windows,
Linux, and OSX.


jina:
cool, thanks @TheJaredWilcurt. next time the site gets deployed, this will be live. :)

PR #179, opened 2016-09-17
Wrapped Haml files at 80 characters

nschonni:
Since the CI system yells about this when you modify them going forward

### DIFF
--- a/source/about.html.haml
+++ b/source/about.html.haml
@@ -1,22 +1,29 @@
 %h1 About Sass
 
 %p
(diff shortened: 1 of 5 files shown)


jina:
Super appreciate you doing this! In the code style guide, we previously ask for 78 characters (http://sass-lang.com/styleguide/code)… but I'll change it to 80.

For the backstory, in the rush to get the site launched, lots of contributors jumped in and content was written super quickly to get the site launched in time for the first Sass Conf. :) The code style wasn't enforced then. But now with HoundCI in place, it calls it out. :)

Thanks again!

PR #184, opened 2017-02-14
Fix broken link to Compass.app site

mike-north:
Looks like the site has been moved to a different domain


mike-north:
This is redundant with #180

PR #189, opened 2017-07-05
Fix misindentation on import code snippet for scss

LuisBoehme:
I was reading the documentation, and i noticed some "misidentation" in the scss code. Thats it, thanks for the awesome project!


jina:
It was actually intentional (some folks like to line up at the colons) but I'll go ahead and merge this.

Issue #192, opened 2017-10-23
Sass Basics: demonstrate extend with placeholder instead of class


ashleykolodziej:
Extending classes in Sass can lead to unexpected results in your compiled CSS if you're not familiar with how extending works, and you've happened to nest that class or element elsewhere in the Sass file in the process of writing other styles. This blog explains the issue that comes up with extending classes that are nested well: https://tiffanybbrown.com/2017/03/dont-use-extend-sass/index.html

I think the best solution is to adjust the extend/inheritance example in the [Sass Guide](http://sass-lang.com/guide) to use a placeholder instead of a class, which behaves in a safer and more expected way when compiling, and link to the [documentation on the extend directive](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_extend__extend) for more information. We could even mention there that CSS classes are also extendable, but may have unintended consequences when the selectors are generated.

I'm happy to do a pull request for this if you're interested!

jina:
Yes, could you please do a pull request? Thanks!

PR #193, opened 2017-10-24
Use placeholders in extend documentation

ashleykolodziej:
Fixes https://github.com/sass/sass-site/issues/192 by updating extend documentation to use placeholder classes in the example Sass/SCSS, and explaining how extend works along with the benefits.


ashleykolodziej:
Good idea, added 👍

jina:
thanks again! :)

PR #194, opened 2017-10-25
Clarifies directory syntax in sass command

ashleykolodziej:
Fixes https://github.com/sass/sass-site/issues/190 with clearer instructions on `sass` and the `--watch` flag, intros the concept of watching for people who might not have worked with preprocessors before, and clarifies the syntax for building/watching to and from directories.


jina:
thank you!

PR #198, opened 2017-12-27
Update install.html.haml

vidjuheffex:
Had trouble installing. This one sentence would have saved me a half an hour.


jina:
@chriseppstein or @nex3 can you confirm this PR is accurate? I am not knowledgeable about this. :)

nex3:
Yeah, this is correct. `sass` depends on `listen`, which requires native extensions, which require `build-essential`.

jina:
great, thanks!

PR #201, opened 2018-02-03
Add sass.cr to the list of LibSass wrappers

GloverDonovan:
This pull request adds [sass.cr](https://github.com/straight-shoota/sass.cr) to the list of LibSass wrappers. I also fixed some typos as I found them.


straight-shoota:
Is this addition welcome?

xzyfer:
This seems reasonable to me. We made a similar addition to the LibSass docs.

jina:
It was welcome. I just wasn't knowledgeable enough about it to make the call. Thanks for taking a look, @xzyfer. :)

Issue #203, opened 2018-03-06
Setup deployment from CI


nschonni:
I think I recall seeing the site is hosted on Heroku, so maybe Travis-CI could be used https://docs.travis-ci.com/user/deployment/heroku/
The maintainers can still keep the deploy key encrypted so it doesn't leak https://docs.travis-ci.com/user/encryption-keys/

jina:
If it doesn't make things harder to maintain, I'm ok with it. I'll let @nex3 decide. Do you have the bandwidth to set it up? I'd also love if our deploys were automated because I have trouble with them sometimes. :smile:

nex3:
I'll see if I can get something working here.

nex3:
This should be good to go! It turns out we don't need Travis at all—Heroku natively supports deploying from GitHub. Just push to `master` and Heroku will take care of the rest.

jina:
wow! that's awesome. :)

PR #204, opened 2018-03-09
Automatically deploy when pushing to Heroku

nex3:
This is a partial step towards #203. I also want to set up Travis to continuously deploy to Heroku every time master updates, but that'll come later.

You can pretty much ignore all the changes in `yard/`, they're just a copy of the old default YARD template (since the new one is worse for reading Markdown files, which is what most of our users want).


jina:
Oooh. Exciting! I'm not the best person to review any Ruby code, but I love the idea of making deploying better so thumbs up from me! :)

nex3:
Is there another reviewer I should add, or am I good to land this?

jina:
I pinged the slack to see if someone wanted to take a look but I'm guessing it's most likely good to go. :)

PR #260, opened 2018-11-29
Add a table of contents to select documentation pages

nex3:
This seems to hang the server for some pages, for reasons I still need
to figure out.


nex3:
I can't reproduce the hangs anymore :woman_shrugging:.

PR #283, opened 2019-02-02
Update the brew command to install sass with one command

dutchmartin:
The brew command on the install page does not work and will result in a error that will ask new users to tap it first.

image

Adding the tap first and then installing sass will work, but the currrent path is not optimal.
Therefore i suggest changing the command.

### DIFF
--- a/source/install.html.haml
+++ b/source/install.html.haml
@@ -122,4 +122,4 @@ title: Install Sass
 
         %pre
           :preserve
-            brew install sass/sass/sass
+            brew tap sass/sass && brew install sass/sass/sass


mistydemeo:
This should be supported - it works for me using the current version of Homebrew. If you can `brew rm dart sass && brew untap sass/sass`, can you get me some debugging info with `brew install --debug sass/sass/sass` so I can figure out what's going on?

dutchmartin:
@mistydemeo Tried it today and it works somehow with the latest version, maybe I have upgraded my homebrew in the last month, or the homebrew taps repo has had some changes.
Anyhow, this pr is not relevant anymore.

PR #284, opened 2019-02-06
Stop promoting Compass

PowerKiKi:
Because the Compass project has been declared deprecated almost 3 years ago
in https://github.com/Compass/compass/commit/dd74a1cfef478a896d03152b2c2a3b93839d168e,
we should stop promoting it as a viable solution.


PowerKiKi:
Travis failure seems unrelated to this change. Other PRs have same, or similar, failures. It seems that htmlproofer made tests flaky.

jina:
This is already going away in the next site release. Thanks though!

PR #294, opened 2019-03-04
Fix watch command

kephas:
The current line was refused by `sass` v3.7.3 but the corrected one worked.


nex3:
You're using [Ruby Sass](https://sass-lang.com/ruby-sass), which is deprecated. The guide is written for [Dart Sass](https://sass-lang.com/dart-sass), which is the new reference implementation.

PR #301, opened 2019-03-05
Include return types in function signatures

### LINKED ISSUE #298: Put return types in function signatures
Return types should look like they're comments written after function signatures, like `//=> color` or `// returns color` (with the type name linked, of course).

nex3:
Closes #298


jina:
this one has some conflicts, too. ill come back to this once those get resolved.

PR #314, opened 2019-03-28
[docs] Sticky Navigation

### LINKED ISSUE #268: make navs sticky

adekunleoduye:
**Summary:**
Documentation nav becomes sticky when user scrolls down the page. 

**Screenshot:**
15126189ffbfee2d47a0ce7b259d557e

To dos:
- [ ] Removes inline styles from `sass.js` and put them into a sass files
- [ ] abstract function out of `sass.js` into `documentation.js`
- [ ] Give background color to nav
- [ ] Add shim to content when nav is sticky

Closes #268

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
-ruby "2.5.1"
+ruby "2.5.3"
 
 gem "builder",                "~> 3.2.2"
 gem "html-proofer",           "~> 3.10"
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
-ruby "2.5.3"
+ruby "2.5.1"
 
 gem "builder",                "~> 3.2.2"
 gem "html-proofer",           "~> 3.10"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -207,7 +207,7 @@ DEPENDENCIES
   yard (~> 0.9)
 
 RUBY VERSION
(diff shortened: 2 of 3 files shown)


jina:
It looks like the content jumps to the left, assuming because the nav is pulled out of the flow. Perhaps a shim of some sort in its place so that the content doesn't shift over?

adekunleoduye:
@jina That sounds good to me. I'll added that to the todo list and update that PR.

nex3:
When I load this up and go to `/documentation`, when I scroll down far enough it just jumps back to the top of the page in a very jarring way.

When I go to `/documentation/syntax/structure` and scroll down, the page nav starts overlapping with the global nav:

nav

Are there plans for making the top-level nav sections expandable? What about for making page-level navs sticky?

I think it may be cleaner to just have the scrollbar present all the time, rather than flashing into existence once the user scrolls down far enough.

Gemfile
```diff
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
-ruby "2.5.3"
+ruby "2.5.1"
```
d2s: Is there a good reason for why the `ruby "2.5.3"` version was replaced with an older `ruby "2.5.1"` version?

adekunleoduye: No specific reason, I was having trouble installing `ruby "2.5.3"`

adekunleoduye:
Since this PR is stale, I'm going to close it and open a new one.

PR #316, opened 2019-03-28
Add special styling for intro pages

### LINKED ISSUE #276: Make intro pages look better
I apologize for the particularly subjective nature of this issue.

There are a handful of reference pages that mostly serve as landing pages to direct users to another page that answers whatever question they may have. Examples include /documentation/at-rules, /documentation/values, /documentation/operators, and most prominently /documentation itself. Right now, these pages feel kind of *unprofessional* to me. I can't put my finger on exactly why... maybe it's because they have so little structure that they're reminiscent of raw HTML, maybe it's because the text takes up relatively little of the screen, I'm not sure. But I think it's important that they look as slick as the rest of the documentation in order to make a good impression on new readers.

I can easily add a class to these pages so that they can get special styling, but I don't know what that styling should be. My best guesses are to bump up the font size and maybe add some special list styling (since all of them have lists), but I trust your design judgment above my own.

nex3:
This just bumps up the font size a bit, to help emphasize that these
pages are special and make them look less bare.

Closes #276

### DIFF
--- a/source/assets/css/components/_index.scss
+++ b/source/assets/css/components/_index.scss
@@ -1,0 +1,7 @@
+@include sl-breakpoint--medium {
+  .sl-c-index {
+    font-size: 1.3rem;
+
+    .sl-c-callout { font-size: 1rem; }
+  }
+}
--- a/source/assets/css/components/_index.scss
+++ b/source/assets/css/components/_index.scss
@@ -1,7 +1,0 @@
-@include sl-breakpoint--medium {
-  .sl-c-index {
-    font-size: 1.3rem;
-
-    .sl-c-callout { font-size: 1rem; }
-  }
-}
--- a/source/layouts/has_both_sidebars.haml
+++ b/source/layouts/has_both_sidebars.haml
@@ -14,7 +14,7 @@
 
             .sl-l-large-holy-grail__main.docSearch-content
               - container = current_page.data.no_container ? '' : 'sl-l-container sl-l-container--small'
-              - container << ' sl-c-index' if current_page.data.intro
+              - container << ' sl-l-container--overview' if current_page.data.intro
               %div{class: container}
                 - if current_page.data.introduction
                   - center_introduction = current_page.data.center_introduction ? 'sl-c-introduction--center' : ''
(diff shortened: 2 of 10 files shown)


source/assets/css/components/_index.scss
```diff
@@ -0,0 +1,7 @@
+@include sl-breakpoint--medium {
+  .sl-c-index {
+    font-size: 1.3rem;
```
jina: This comes out to `20.8` rendered pixels. Can we make it `1.25rem` so it comes out to an even 20?

source/assets/css/components/_index.scss
```diff
@@ -0,0 +1,7 @@
+@include sl-breakpoint--medium {
```
jina: Can this go in the components/callouts file instead?

source/layouts/has_both_sidebars.haml
```diff
@@ -14,6 +14,7 @@
 
             .sl-l-large-holy-grail__main.docSearch-content
               - container = current_page.data.no_container ? '' : 'sl-l-container sl-l-container--small'
+              - container << ' sl-c-index' if current_page.data.intro
```
jina: `sl-c-` prefix means component. The `index` here is not really a component in itself.

I think I'd rather see this as a modifier on the intro itself since all it does it modify it's size.

So it would have both these classes:
so: `sl-c-introduction sl-c-introduction--large`.

And then the CSS changed to not be a nested style but a modified style.

source/assets/css/components/_index.scss
nex3: Done (although this mostly doesn't have to do with callouts).

source/assets/css/components/_index.scss
nex3: Done. My intutition is that 20px is too small and 21px is too big, though, so I encourage you to reconsider.

source/layouts/has_both_sidebars.haml
nex3: > `sl-c-` prefix means component. The `index` here is not really a component in itself.

Got it, that makes sense.

> I think I'd rather see this as a modifier on the intro itself since all it does it modify it's size.
> 
> So it would have both these classes:
> so: `sl-c-introduction sl-c-introduction--large`.
> 
> And then the CSS changed to not be a nested style but a modified style.

I think my choice of names may have been confusing here... the "intro" field I'm adding is different than the "introduction" field you added earlier. It's meant to serve as a flag that the entire page serves as an introduction, and so the whole thing should be bigger. I've renamed the field to "overview" to make that more clear, and changed the class to be `sl-l-container--overview` to make it clear that it's modifying the layout.

nex3:
Landing this TBR

PR #317, opened 2019-03-29
Render syntax examples using adjacent tabs if they're narrow enough

### LINKED ISSUE #311: Show input and output of examples side by side
It would be nice to be able to see the Sass/SCSS source and the CSS output side-by-side when viewing examples, at least when the viewport is wide enough to fit both.

Screenshot 2019-03-19 at 1 40 31 PM

nex3:
Closes #311


nex3:
Landing this TBR

PR #319, opened 2019-04-05
Indicate that Ruby Sass is end-of-life

### LINKED ISSUE #206: Remove references to Ruby Sass
**Note:** This should only happen once Ruby Sass reaches its end-of-life, a year after Dart Sass's first stable release.

We should get rid of most references to Ruby Sass once it's no longer a going concern. It probably merits a mention somewhere as "this is an old thing that used to be used", but it should no longer be listed as a main-line implementation.

nex3:
Closes #206

### DIFF
--- a/source/ruby-sass.html.haml
+++ b/source/ruby-sass.html.haml
@@ -25,9 +25,9 @@ title: Ruby Sass
       has faded into the background. At the same time, Sass projects have grown
       far larger than we initially envisioned them, and their performance needs
       have outpaced the speed Ruby can provide. Both [Dart Sass][] and
-      [LibSass][] are blazing fast and easy to install, especially in Node. Ruby
-      Sass couldn't keep up, and it didn't make sense to spend the core team's
-      resources on it any longer.
+      [LibSass][] are blazing fast, easy to install, and are readily available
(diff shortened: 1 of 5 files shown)


source/ruby-sass.html.haml
```diff
@@ -3,51 +3,35 @@ title: Ruby Sass
...
 
-      Support for Ruby Sass is being turned down because the Sass core team no
-      longer has time to maintain it as well as [Dart Sass][] and [LibSass][].
-      But that doesn't mean it has to go away. If a community member wanted to
-      take on the task of maintaining it, the core team would be happy to help
-      get them up to speed. It's not an easy job: it would require keeping on
-      top of bug fixes and implementing new language features as they're
-      designed by the language team. But it would also be an opportunity to work
-      on a project that's used by thousands of people every day.
+      Since then, Node.js has become ubiquitous for frontend tooling while Ruby
```
srawlins: Since we're mentioning Node here, does it make sense to remark that the Dart implementation makes itself available as a Node.js module, since JavaScript is a first-class compilation target for Dart?

nex3: Done.

Issue #321, opened 2019-04-06
Import section in the guide page have typo errors


Ganofins:
Import section in the guide page have some type errors at 
Let's say you have a couple of Sass files, _reset.scss and base.scss. We want to import _reset.scss into base.scss

so in the above paragraph it is mentioned couple of sass files but the extension to those files is scss

nex3:
"Sass" is used to refer to the language as a whole, including both SCSS and the indented syntax. So when the documentation says "Sass files", it means any kind of Sass file; it would work just as well with files in the indented syntax.

Ganofins:
Thanks for letting me know

PR #322, opened 2019-04-06
fixed the typo error of sass

Ganofins:
#321 I have fixed that typo error of sass to SCSS


nex3:
See https://github.com/sass/sass-site/issues/321#issuecomment-480633961.

PR #337, opened 2019-05-03
Remove duplicate session about mixins

aulisius:
The section about mixins in the Guide page was repeated twice. I removed the duplicate that was below the Operators topic.

image


nex3:
Thanks for the fix!

PR #340, opened 2019-05-08
Bundle all JS into a single file

nex3:
Because we weren't doing this yet, we weren't actually loading the
redirect.js JS code at all.

This also gets rid of some unused JS.


nex3:
AFAIK no one's ever really bothered to configure the linter for the code style that's actually used in the project, so I mostly just ignore it.

PR #342, opened 2019-05-15
More clearly document import order

nex3:
See sass/dart-sass#574

### DIFF
--- a/source/documentation/js-api.html.md.erb
+++ b/source/documentation/js-api.html.md.erb
@@ -727,8 +727,9 @@ h1 {
 <% end %>
 
 <% impl_status dart: '1.20.2', node: false, feature: "Import order" do %>
-  Versions of Dart Sass before 1.20.2 loaded imports from load paths if possible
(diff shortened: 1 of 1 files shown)


source/documentation/js-api.html.md.erb
```diff
@@ -726,6 +726,20 @@ h1 {
   Versions of Node Sass before 2.0.0 don't support the `importer` option at all.
 <% end %>
 
+<% impl_status dart: '1.20.2', node: false, feature: "Import order" do %>
+  Versions of Dart Sass before 1.20.2 loaded imports from load paths if possible
+  before passing them to importers.
```
jelbourn: What do you think about the phrasing
```markdown
Prior to v1.20.2, Dart Sass preferred resolving imports via load paths ((`includePaths`))[#includePaths] before using custom importers.
```

nex3: Done, mostly.

PR #344, opened 2019-05-22
Update guide.html.haml

jrl05k:
```
sass --watch input.scss output.css
``` 
should be 
```
sass --watch input.scss:output.css
```


nex3:
Both of these invocations are valid for Dart Sass. The colon-separated syntax is only necessary when you're compiling multiple files or directories at once.

PR #355, opened 2019-07-19
Fixes correct version of libsass for index imports

fetis:
This feature actually implemented only in libsass 3.6.0
https://github.com/sass/libsass/releases/tag/3.6.0

here's the commit `[3af837c] Add support for index imports (@xzyfer)`


nex3:
Thanks for the fix!

PR #356, opened 2019-08-06
Add an implementation note about sass/libsass#2957

nex3:


source/documentation/at-rules/import.html.md.erb
```diff
@@ -361,6 +361,16 @@ rendered as-is. It can even be [extended][]!
 
 ## Plain CSS `@import`s
 
+<% impl_status dart: true, libsass: :partial, ruby: true do %>
+  By default, LibSass handles plain CSS imports correctly. However, if [custom
+  importers][] are passed to LibSass, they will incorrectly be invoked for
+  imports that should be plain-CSS, making it possible for those imports to load
+  Sass files.
```
jelbourn:
```markdown
By default, LibSass handles plain CSS imports correctly. However, any
[custom importers][] will incorrectly apply to plain-CSS import statements, which allows
those import statements to load Sass files.
```
(more active)

nex3: Done.

PR #357, opened 2019-08-14
Sassify Sass Syntax

TheJaredWilcurt:


nex3:
Thanks for the contribution! However, while we're glad that the shorthand `=` and `+` syntax is available for advanced users who highly value terseness, we feel that they're generally more difficult to understand at a glance for newer users and so we'd like to keep them out of the documentation examples.

PR #368, opened 2019-09-05
Document @forward

### LINKED ISSUE #362: Document @forward
The new `@forward` rule needs to be documented in the module-ified documentation.

nex3:
Closes #362

### DIFF
--- a/source/documentation/at-rules/forward.html.md.erb
+++ b/source/documentation/at-rules/forward.html.md.erb
@@ -1,25 +1,27 @@
 ---
 title: "@forward"
 introduction: >
-  The `@forward` rule makes [mixins](mixin), [functions](function), and
-  [variables](../variables) from another Sass stylesheet available to
-  stylesheets that load your stylesheet with the [`@use` rule](use). It makes it
-  possible to organize Sass libraries across many files, while allowing their
-  users to load a single entrypoint file.
+  The `@forward` rule loads a Sass stylesheet and makes its [mixins](mixin),
+  [functions](function), and [variables](../variables) available when your
+  stylesheet is loaded with the [`@use` rule](use). It makes it possible to
+  organize Sass libraries across many files, while allowing their users to load
+  a single entrypoint file.
 ---
 
-It's written `@forward "<url>"`. It loads the module at the given URL just like
-`@use`, but it makes the [public][] members of the loaded module available to
-users of your module as though they were defined directly in your module. Those
-members aren't visible in your module, though—if you want that, you'll need to
-write a `@use` rule as well. Don't worry, it'll only load the module once!
+The rule is written `@forward "<url>"`. It loads the module at the given URL
+just like `@use`, but it makes the [public][] members of the loaded module
+available to users of your module as though they were defined directly in your
+module. Those members aren't available in your module, though—if you want that,
+you'll need to write a `@use` rule as well. Don't worry, it'll only load the
+module once!
 
 [public]: use#private-members
 
 <% fun_fact do %>
   The `@forward` rule acts just like `@use` when it comes to a module's CSS.
   Styles from a forwarded module will be included in the compiled CSS output,
-  and your module can [extend][] it, even if it isn't also `@use`d.
+  and the module with the `@forward` can [extend][] it, even if it isn't also
+  `@use`d.
 
   [extend]: extend
 <% end %>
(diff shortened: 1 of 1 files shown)


source/documentation/at-rules/forward.html.md.erb
```diff
@@ -1,7 +1,179 @@
 ---
 title: "@forward"
 introduction: >
-  TODO(nweiz): Write this
+  The `@forward` rule makes [mixins](mixin), [functions](function), and
+  [variables](../variables) from another Sass stylesheet available to
+  stylesheets that load your stylesheet with the [`@use` rule](use). It makes it
+  possible to organize Sass libraries across many files, while allowing their
+  users to load a single entrypoint file.
```
jelbourn: Using the word "stylesheet" three times so close together makes the sentence a bit hard to read, but I'm also having a hard time with an alternate wording. I'm thinking something kind of like 
```markdown
The `@forward` rule makes mixins, functions, and variables available for `@use` from a different
stylesheet than the one in which they're declared. A stylesheet using this rule _forwards_ these
symbols from the original location through itself. This allows Sass libraries to surface APIs from
multiple files through a single entry-point.
```

I'm also wondering if it makes sense to add a comparison like "The process of forwarding works similarly to
re-exporting symbols in JavaScript, Python, etc."

source/documentation/at-rules/forward.html.md.erb
```diff
@@ -1,7 +1,179 @@
...
 ---
 
+It's written `@forward "<url>"`. It loads the module at the given URL just like
```
jelbourn: "It's written" -> "The rule uses the syntax"

("It" here could be ambiguous, plus a more active voice, though it is longer)

source/documentation/at-rules/forward.html.md.erb
```diff
@@ -1,7 +1,179 @@
...
+`@use`, but it makes the [public][] members of the loaded module available to
+users of your module as though they were defined directly in your module. Those
+members aren't visible in your module, though—if you want that, you'll need to
```
jelbourn: "visible" -> "available"?

source/documentation/at-rules/forward.html.md.erb
```diff
@@ -1,7 +1,179 @@
...
+write a `@use` rule as well. Don't worry, it'll only load the module once!
+
+[public]: use#private-members
+
+<% fun_fact do %>
+  The `@forward` rule acts just like `@use` when it comes to a module's CSS.
+  Styles from a forwarded module will be included in the compiled CSS output,
+  and your module can [extend][] it, even if it isn't also `@use`d.
```
jelbourn: Does this apply to the file in which the `@forward` resides as well? It isn't super clear to me, since the _symbols_ from the forwarded file _aren't_ available.

source/documentation/at-rules/forward.html.md.erb
nex3: > Using the word "stylesheet" three times so close together makes the sentence a bit hard to read, but I'm also having a hard time with an alternate wording. I'm thinking something kind of like
> 
> ```gfm
> The `@forward` rule makes mixins, functions, and variables available for `@use` from a different
> stylesheet than the one in which they're declared. A stylesheet using this rule _forwards_ these
> symbols from the original location through itself. This allows Sass libraries to surface APIs from
> multiple files through a single entry-point.
> ```

I don't think I like the "available for use from a different stylesheet than the one in which they're declared", because that could arguably describe `@use` as well.  I've tried rewriting it to make the repetition less awkward; PTAL.

> I'm also wondering if it makes sense to add a comparison like "The process of forwarding works similarly to re-exporting symbols in JavaScript, Python, etc."

In my experience, the people who most struggle to understand these docs are those coming to them from a non-programming background. I expect that most people who are familiar with re-exporting symbols will grok this pretty easily regardless.

source/documentation/at-rules/forward.html.md.erb
nex3: Changed to "The rule is written". I want to keep the word "written" because it's consistently used as a signal word for "here comes the general syntax".

source/documentation/at-rules/forward.html.md.erb
nex3: Done.

source/documentation/at-rules/forward.html.md.erb
nex3: Yes. Clarified.

PR #370, opened 2019-09-11
Document new built-in functions

nex3:


source/documentation/modules/meta.html.md.erb
```diff
@@ -4,6 +4,99 @@ title: sass:meta
...
+## Mixins
+
+<% function 'meta.load-css($url, $with: null)' do %>
+  <% impl_status dart: '(unreleased)', libsass: false, ruby: false do %>
+    Only Dart Sass currently supports this mixin.
+  <% end %>
+
+  Loads the [module][] at `$url` and includes its CSS as though it were written
+  as the contents of this mixin. The `$with` parameter provides
+  [configuration][] for the modules; if it's passed, it must be a map from
```
jelbourn: "it's passed" -> "provided" ?

source/documentation/modules/meta.html.md.erb
```diff
@@ -4,6 +4,99 @@ title: sass:meta
...
+  variable names (without `$`) to the values of those variables to use in the
+  loaded module.
```
jelbourn: Does the map have to have _all_ the variables, or can it be a subset?

source/documentation/modules/meta.html.md.erb
```diff
@@ -198,6 +291,20 @@ title: sass:meta
 <% end %>
 
 
+<% function 'meta.keywords($args)', 'keywords($args)', returns: 'map' do %>
```
jelbourn: Just curious: why "keywords" instead of something like "argument-map"?

source/documentation/modules/meta.html.md.erb
```diff
@@ -4,6 +4,99 @@ title: sass:meta
...
+    ---
+    // style.scss
+    @use "sass:meta";
+
+    body.dark {
+      @include meta.load-css("dark-theme/code",
+          $with: ("border-contrast": true));
+    }
+    ===
+    // dark-theme/_code.scss
```
Awjin: This should be `.sass`

source/documentation/modules/meta.html.md.erb
```diff
@@ -4,6 +4,99 @@ title: sass:meta
...
+    $border-contrast: false !default
+
+    code
+      background-color: #6b717f
+      color: #d2e1dd
+      @if $border-contrast
+        border-color: #dadbdf
+    ---
+    // style.scss
```
Awjin: `.sass`

source/documentation/modules/meta.html.md.erb
nex3: Good catch! Done.

source/documentation/modules/meta.html.md.erb
nex3: Done.

PR #371, opened 2019-09-11
Update the repository for the PHP binding for libsass

stof:
The https://github.com/sensational/sassphp repo is not updated anymore. On the other hand, https://github.com/absalomedia/sassphp is the source used to package this binding in Ubuntu 18.04 and newer.

### DIFF
--- a/source/libsass.html.haml
+++ b/source/libsass.html.haml
@@ -120,7 +120,7 @@ introduction: >
     :markdown
       ### PHP
 
-      The [SassPHP](https://github.com/sensational/sassphp) project is an
+      The [SassPHP](https://github.com/absalomedia/sassphp) project is an
       updated fork of an [older PHP
       version](https://github.com/jamierumbelow/sassphp).
 


nex3:
@sensational Can you confirm? It looks like https://github.com/sensational/sassphp was updated *more recently* than https://github.com/absalomedia/sassphp (17 July vs 1 July).

stof:
hmm, it looks like the sensational repo was revived in July, finally adding PHP 7 support (which was released in December 2015).
On the other hand, the absalomedia repo supports PHP 7 since a very long time, and even tried enabling CI on the upcoming PHP 7.4 (not yet compatible). The last update at the beginning of July is because the work happening there seems mostly tied to libsass upgrades, and there has been no libsass releases since then.

hudri:
I'm with @stof in this, sensational is still von libsass v3.6.1, while absalomedia already has updated to libsass v.3.6.3. Absalomedia seems to maintain and update their repo frequently, they support PHP up to v7.4 and already have tests for PHP v8

nex3:
Since we haven't heard back from @sensational at all, I think it's safe to say their project is outdated.

PR #384, opened 2019-10-10
Recommend that users write @forward before @use

nex3:


nex3:
I'd lean towards 1. @jelbourn what do you think?

PR #406, opened 2019-11-07
Temporarily roll back sticky navigation until we can fix #400

### LINKED ISSUE #400: guide page scss/sass toggle doesn't work
https://sass-lang.com/guide

just jumps down the page when trying to toggle between sass and scss examples

nex3:


nex3:
It looks like this isn't actually solving the problem, so I'm guessing sticky navs isn't the root cause. Maybe it's #398? I'm going to try rolling that back instead.

PR #408, opened 2019-11-08
[hot-fix] Rollback sticky nav

adekunleoduye:
Rolling back sticky nav because of sass/scss toggle issue (#400).


nex3:
See https://github.com/sass/sass-site/pull/406; I don't think sticky nav is the root cause here.

PR #420, opened 2019-12-10
Typos fixed on the modules/selector page

alexander-doroshko:
Typos fixed on the modules/selector page:
`selector.unify()` -> `selector-unify()`, 
`compound-selectors()` -> `simple-selectors()`


Awjin:
Good catch, thanks!

PR #421, opened 2019-12-13
Accessibility fixes

tatianamac:
Hello! 👋🏽 

I've made some accessibility fixes to the site. Principally, I:
- Added a skip link so keyboard users can skip navigation (New design element, so added @jina per contribution guidelines)
- Made colour contrast ratios to WCAG AA standards. Stayed within style guide colour palette.
- Converted `<asides>` into `<divs>` to fix landmark confusion. The `<asides>` can be added back later, but there's a lot of `<divs>` scaffolding it that I didn't feel comfortable removing it.
- Ensured `<h1>...<h6>` were used properly and in order.
- Ensured all elements were within landmark tags.
- Removed redundant `aria-roles`, `aria-labels`, etc. Thank u, next, ARIA.
- Removed empty `<a>`, which was causing confusing content order/keyboard experience.
- Enhanced keyboard accessibility for menus and tabs.

**Outstanding things I need help on:**
- `<dl>` are nested with extra `<div>` scaffolding, which throws semantic errors. I can't figure out how to remove these extra divs without destroying the layout.
- The SCSS/SASS/CSS toggle on the Learn Sass page should become a `role=tablist` in my opinion, but there appears to be some JS that controls this tabbing behaviour. The headers are also listed as `role=presentation` which I disagree with. I would feel more comfortable if I gained better understanding of the intent here and received assistance implementing this fix.
- Certain `<a>` tags seem to inherit `tabindex="-1"` and role="presentation"`, which causes a conflict (as all `<a>` should be accessible). I can't find where this is being auto-generated from.
- The Rouge syntax highlighter intrinsically has an inaccessible palette. Fixing the base palette here will solve dozens of contrast issues.
- The search field should have a label and better input control. However, as it's relying on Aglolia for a lot of its focus management, I did not want to full-suite rewrite the input styling.
- Blog section? I audited while it was still on Logdown. I went to the PR with that but it sends me to this [Heruko error page](https://sass-lang-pr-409.herokuapp.com/)—I'm new so maybe I am doing something wrong here?

Thank you for everyone's patience with me and for your help!


nex3:
I'm still on vacation so I don't have time for a full review, but I'll try to address the questions:

> * `<dl>` are nested with extra `<div>` scaffolding, which throws semantic errors. I can't figure out how to remove these extra divs without destroying the layout.

This is beyond my expertise--@jina, can you provide some guidance?

> * The SCSS/SASS/CSS toggle on the Learn Sass page should become a `role=tablist` in my opinion, but there appears to be some JS that controls this tabbing behaviour. The headers are also listed as `role=presentation` which I disagree with. I would feel more comfortable if I gained better understanding of the intent here and received assistance implementing this fix.

That's all handled by the [jQuery UI tabs](https://jqueryui.com/tabs/) plugin, which is invoked [here](https://github.com/sass/sass-site/blob/master/source/assets/js/components/sass-syntax-switcher.js). Looking at it in the dev inspector, though, it looks like it does have a `role=tablist`, although I agree that `role=presentation` seems wrong for the individual headers. I'm not sure what the best way is to customize this plugin.

> * Certain `<a>` tags seem to inherit `tabindex="-1"` and role="presentation"`, which causes a conflict (as all `` should be accessible). I can't find where this is being auto-generated from.

Are these the same `a` tags that are in the tab headers?

> * The Rouge syntax highlighter intrinsically has an inaccessible palette. Fixing the base palette here will solve dozens of contrast issues.

It looks like the theme is currently [generated from Rogue](https://github.com/sass/sass-site/blob/70075be617297d5b5f94315b8716504396ca8e4a/source/assets/css/vendor/_syntax.scss.erb), but it doesn't have to be... would it be possible for you to copy it into the stylesheet directly and modify as appropriate to make the palette accessible?

> * The search field should have a label and better input control. However, as it's relying on Aglolia for a lot of its focus management, I did not want to full-suite rewrite the input styling.

Is there some low-hanging fruit we can do here without diving into Algolia code, such as adding a label?

> * Blog section? I audited while it was still on Logdown. I went to the PR with that but it sends me to this [Heruko error page](https://sass-lang-pr-409.herokuapp.com/)—I'm new so maybe I am doing something wrong here?

Oh yeah, the Heroku previews expire after a certain amount of time. Have you been able to get the site up and running locally? That's the most reliable way to view it. I've also renewed the review app, so it should be accessible again for a while.

source/assets/css/components/_alerts.scss
```diff
@@ -5,7 +5,7 @@
   };
   text-align: center;
   background: darken($sl-color--hopbush, 10%);
-  color: rgba($sl-color--white, .875);
+  color: rgba($sl-color--white, 1);
```
jina: If we are gonna make this `1`, then we need to:

1. Not have this wrapped in the `rgba` function as it is pointless now. :)
2. Consider if the border on the links is enough now to differentiate them from the content, now that they are the same color.

source/assets/css/regions/_banner.scss
```diff
@@ -3,10 +3,15 @@
 
   a {
     display: inline-block;
+  }
+
+  &__brand
+  
+  a {
```
jina: Should this `a` be nested (meaning a curly brace is missing) or is it meant to be a part of the `&__brand` selector? If so, we should bring that up to the same line.

source/assets/css/visual-design/_theme.scss
```diff
@@ -328,7 +328,7 @@ html {
 }
 
 body {
-  background: transparent;
+  background: white;
```
jina: This is going to cover up the background texture on the `html` element, isn't it?

source/assets/css/components/_navigation.scss
```diff
@@ -8,3 +8,21 @@
...
+  padding: 0.25em 1em;
+  transform: translateY(calc(-100% - 1em));
+  position: absolute;
+  left: 1em;
+  top: 1em;
+  z-index: 9999;
+
+  &:focus {
+    transform: translateY(0);
+    transition: all .1s ease-in-out;
```
Awjin: Could this transition be on the parent instead, so that the link's exit is also animated?

source/assets/js/components/list-navigation.js
```diff
@@ -2,7 +2,7 @@ $(function() {
   $(".sl-c-list-navigation-wrapper--collapsible li > ul")
       .parent()
       .children("a")
-      .removeAttr("href")
+      // .removeAttr("href") Remove attr was causing keyboard focus skips
```
Awjin: This can be deleted.

source/layouts/layout.haml
```diff
@@ -46,24 +46,26 @@
...
+      Skip to content
+        
     %header.sl-r-banner(itemtype='http://schema.org/WPHeader' itemscope='itemscope' role='banner')
+      .sl-c-alert.sl-c-alert--info
+        .sl-l-container
+          %p
+            %strong Sass just launched a brand new module system.
+            = succeed '!' do
+              = link_to '/blog/posts/7858341-the-module-system-is-launched' do
+                Learn all about the new on the Sass blog
```
Awjin: This should remain as "Learn all about it on the Sass blog".

Awjin:
> > * The SCSS/SASS/CSS toggle on the Learn Sass page should become a `role=tablist` in my opinion, but there appears to be some JS that controls this tabbing behaviour. The headers are also listed as `role=presentation` which I disagree with. I would feel more comfortable if I gained better understanding of the intent here and received assistance implementing this fix.
> 
> That's all handled by the [jQuery UI tabs](https://jqueryui.com/tabs/) plugin, which is invoked [here](https://github.com/sass/sass-site/blob/master/source/assets/js/components/sass-syntax-switcher.js). Looking at it in the dev inspector, though, it looks like it does have a `role=tablist`, although I agree that `role=presentation` seems wrong for the individual headers. I'm not sure what the best way is to customize this plugin.

Worst-case, we could use JS to manually replace the `roles` in whatever html the jQuery plugin outputs. Happy to help here if needed.

source/assets/css/visual-design/_theme.scss
tatianamac: Oops! Thank you for catching this. This should be removed; I was doing it to make it easier to auto-run contrast checks. 🙃

jina: Let's keep the transparent there in case someone has user styles defined.

source/layouts/layout.haml
tatianamac: I can revert it, but to provide context: I changed it so that if someone gathers all links all links from the page (which is a common thing that screen reader users do), it's not clear what they're learning.

Let me know how you'd like me to proceed.

config.rb
```diff
@@ -110,7 +110,7 @@
   content.gsub(%r{^<(h[0-6])(.*?)</\1>}m) do |header_text|
     header = Nokogiri::HTML::DocumentFragment.parse(header_text).children.first
     id = header.attr(:id)
-    header.children.before("<a class='anchor' href='##{id}'></a>") if id
+    # header.children.before("<a class='anchor' href='##{id}'></a>") if id // Commenting this out, as it appears to serve no function and is generating empty links, which causes accessibility issues.
```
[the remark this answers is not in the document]
tatianamac: Thanks, @Awjin. The impact is that it is generating empty links, which create confusion for screenreader users. Perhaps we can visually hide the link text using a `<span class="visuallyhidden">` like this:
```
     header.children.before("<a class='anchor' href='##{id}'><span class='visuallyhidden'>{id}</span></a>") if id 
```
Does that work for you?

source/assets/css/components/_alerts.scss
tatianamac: 1. D'oh, yes, fixed—thanks for catching that. 😅
2. I removed the strong tag for better differentiation. How do you feel about this?
Unbolded first sentence

source/assets/css/visual-design/_theme.scss
tatianamac: Yep! Updated it in the latest commit. 💙

tatianamac:
> Ah, it's so awesome to have you on this, Tatiana. I'm giddy. :)
> 
> I'm on a slow wifi on a flight so having trouble pulling this down to look at it, so reviewing based on your description and code here, only. I can submit this review and then do a review again after this round to actually see it in the browser after I'm home.
> 
> Anyway, I left some inline comments. Regarding two other things:
> 
>     * **`div` in `dl`**: Oh snap. It was my understanding that divs were acceptable now in `dl`s (see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#Wrapping_name-value_groups_in_%3Cdiv%3E_elements). However, if this is breaking accessibility, then I'm very open to finding a different way to structure this content semantically in a way that makes sense and still allows for the layout.
Ah, thanks. I think this is a newer acceptable standard than what I was working off of (W3C). For now, I think it's fine to keep the `divs`, I'll continue to do more research and ask around for peace of mind.

>     * **`div` instead of `aside`**: Is `aside` really not appropriate for the tips? They are kinda like interrupting the flow to inject some new info that is related but not part of the main block of content, so I thought this was acceptable. Happy to use `div` instead if that is the case.
Totally. _Contextually,_ it's absolutely appropriate. However, the way it's executed nests it in too many divs. It should live at the same level (first-level) as all other landmarks.
If I could get help extracting it to the highest level from someone more familiar with the codebase, that would be dope.

All of your other comments have either been addressed (responded with ❤️), or responded to with request.

tatianamac:
> That's all handled by the [jQuery UI tabs](https://jqueryui.com/tabs/) plugin, which is invoked [here](https://github.com/sass/sass-site/blob/master/source/assets/js/components/sass-syntax-switcher.js). Looking at it in the dev inspector, though, it looks like it does have a `role=tablist`, although I agree that `role=presentation` seems wrong for the individual headers. I'm not sure what the best way is to customize this plugin.

> > * Certain `<a>` tags seem to inherit `tabindex="-1"` and role="presentation"`, which causes a conflict (as all `` should be accessible). I can't find where this is being auto-generated from.

> Are these the same `a` tags that are in the tab headers?
Yes. @Awjin , if you're able to help here per your comment, that'd be great! We should remove the `role="presentation"` from each of the `a` tags.
> 
> > * The Rouge syntax highlighter intrinsically has an inaccessible palette. Fixing the base palette here will solve dozens of contrast issues.
> 
> It looks like the theme is currently [generated from Rogue](https://github.com/sass/sass-site/blob/70075be617297d5b5f94315b8716504396ca8e4a/source/assets/css/vendor/_syntax.scss.erb), but it doesn't have to be... would it be possible for you to copy it into the stylesheet directly and modify as appropriate to make the palette accessible?
Sorry if I am being dense, but would you like me to track down the stylesheet and create a new stylesheet that's referenced here to override it? Some extra direction here would be helpful—thank you.
> > * The search field should have a label and better input control. However, as it's relying on Aglolia for a lot of its focus management, I did not want to full-suite rewrite the input styling.
> 
> Is there some low-hanging fruit we can do here without diving into Algolia code, such as adding a label?
Sure, I'll add that to the next round of edits and see what else I can do without affecting Algolia. 😃
 
> Oh yeah, the Heroku previews expire after a certain amount of time. Have you been able to get the site up and running locally? That's the most reliable way to view it. I've also renewed the review app, so it should be accessible again for a while.
I might be dense here. Currently the local link still goes to Logdown. What directory should I be looking in for the local instance?

tatianamac:
Thank you, @nex3, @jina, and @Awjin for the edits! I pushed another commit that addresses most of the feedback. Most outstanding edits require input, addressed as individual comments. I'll await that feedback. 💙

config.rb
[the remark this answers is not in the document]
Awjin: That works for me.

Just double-checking, the span should be `<span class='visuallyhidden'>`**#{id}**`</span></a>`.

What's the best practice for accessible text on links? Is it ok that that will generate hyphenated text ("loading-members")?

source/layouts/layout.haml
Awjin: I see. In that case, "Learn all about the module system on the Sass blog" would be more accurate.

config.rb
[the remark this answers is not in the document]
tatianamac: Appreciate the support!

Ah, yeah. Ideally, it would be human readable, so it's preferable that it's "Loading members." Is it possible to either generate the name from the existing header (which is already written properly) or strip out the punctuation with a script? Otherwise the syntax is correct.

source/layouts/layout.haml
tatianamac: Excellent. Will update with next PR. 👍🏽

Awjin:
> Certain `<a>` tags seem to inherit `tabindex="-1"` and `role="presentation"`, which causes a conflict (as all should be accessible). 

Correct me if I'm wrong, as I'm out of my depth here, but I think the tablist is implementing the [behavior as intended](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role). Users tab to whichever tab has `tabindex="0"`, and then can toggle between the tabs by using left/right arrows. The non-active tabs should have `tabindex="-1"`.

The `<a>`s that have `role="presentation"` are definitely weird, but I think it's because they're not actually links and are acting as faux buttons.

config.rb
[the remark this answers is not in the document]
Awjin: `#{header.text}` should give you human readable text

tatianamac:
> > Certain `<a>` tags seem to inherit `tabindex="-1"` and `role="presentation"`, which causes a conflict (as all should be accessible).
> 
> Correct me if I'm wrong, as I'm out of my depth here, but I think the tablist is implementing the [behavior as intended](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role). Users tab to whichever tab has `tabindex="0"`, and then can toggle between the tabs by using left/right arrows. The non-active tabs should have `tabindex="-1"`.
> 
> The `<a>`s that have `role="presentation"` are definitely weird, but I think it's because they're not actually links and are acting as faux buttons.

I could be mistaken in how I'm understanding this—so not hoping to be dogmatic with the following but to gain more understanding all around 😅:

Yes. You are correct that the *behaviour* is as intended for `ul` as `role=tablist` (e.g., tab to each `SCSS` label then arrow right/left to get to `SASS` and `CSS`), but two things come up:

1. `role=presentation` should only be associated with things that should not ignored (and thus not announced in SRs) in the accessibility tree. Links `a` should be announced. It would make sense if `role=presentation` were associated in this code example with the `li`, as it's not genuinely a list item but instead serving as a tab. 

2. The following this is dependent on the first, but we're setting the `tabfocus` to the `li` instead of to the `a`. To me, the `a` is the thing that's the tab, and the `li` is the presentation element.
image

To give an example, [WAI-ARIA's core example]( is almost 1:1 to ours):
```
<ul role="tablist">
  <li role="presentation">
    <a role="tab" href="#">Tab 1</a>
  </li>
  <li role="presentation">
    <a role="tab" href="#">Tab 2</a>
  </li>
  <li role="presentation">
    <a role="tab" href="#">Tab 3</a>
  </li>
</ul>
```
I'd recommend we replicate this. Happy to implement it, but I like to gain a full understanding of _why_ things are done, as often there's good reason for unorthodox techniques.
I hope that makes sense, happy to clarify if not. Let me know your thoughts after this?

(Also bc accessibility execution is very subjective, I'm gathering as much research and opinions on the matter is).

Awjin:
Got it, thanks for educating me on this. AFAICT the only reason for the current markup is that it's what the jQuery plugin outputs. Building on the WAI-ARIA example, how would you recommend we handle `tabindex`?

nex3:
> > > * The Rouge syntax highlighter intrinsically has an inaccessible palette. Fixing the base palette here will solve dozens of contrast issues.
> 
> > It looks like the theme is currently [generated from Rogue](https://github.com/sass/sass-site/blob/70075be617297d5b5f94315b8716504396ca8e4a/source/assets/css/vendor/_syntax.scss.erb), but it doesn't have to be... would it be possible for you to copy it into the stylesheet directly and modify as appropriate to make the palette accessible?
> 
> Sorry if I am being dense, but would you like me to track down the stylesheet and create a new stylesheet that's referenced here to override it? Some extra direction here would be helpful—thank you.

<details>
<summary>Here's the full stylesheet:</summary>

```css
.highlight table td { padding: 5px; }
.highlight table pre { margin: 0; }
.highlight .cm {
  color: #999988;
  font-style: italic;
}
.highlight .cp {
  color: #999999;
  font-weight: bold;
}
.highlight .c1 {
  color: #999988;
  font-style: italic;
}
.highlight .cs {
  color: #999999;
  font-weight: bold;
  font-style: italic;
}
.highlight .c, .highlight .cd {
  color: #999988;
  font-style: italic;
}
.highlight .err {
  color: #a61717;
  background-color: #e3d2d2;
}
.highlight .gd {
  color: #000000;
  background-color: #ffdddd;
}
.highlight .ge {
  color: #000000;
  font-style: italic;
}
.highlight .gr {
  color: #aa0000;
}
.highlight .gh {
  color: #999999;
}
.highlight .gi {
  color: #000000;
  background-color: #ddffdd;
}
.highlight .go {
  color: #888888;
}
.highlight .gp {
  color: #555555;
}
.highlight .gs {
  font-weight: bold;
}
.highlight .gu {
  color: #aaaaaa;
}
.highlight .gt {
  color: #aa0000;
}
.highlight .kc {
  color: #000000;
  font-weight: bold;
}
.highlight .kd {
  color: #000000;
  font-weight: bold;
}
.highlight .kn {
  color: #000000;
  font-weight: bold;
}
.highlight .kp {
  color: #000000;
  font-weight: bold;
}
.highlight .kr {
  color: #000000;
  font-weight: bold;
}
.highlight .kt {
  color: #445588;
  font-weight: bold;
}
.highlight .k, .highlight .kv {
  color: #000000;
  font-weight: bold;
}
.highlight .mf {
  color: #009999;
}
.highlight .mh {
  color: #009999;
}
.highlight .il {
  color: #009999;
}
.highlight .mi {
  color: #009999;
}
.highlight .mo {
  color: #009999;
}
.highlight .m, .highlight .mb, .highlight .mx {
  color: #009999;
}
.highlight .sb {
  color: #d14;
}
.highlight .sc {
  color: #d14;
}
.highlight .sd {
  color: #d14;
}
.highlight .s2 {
  color: #d14;
}
.highlight .se {
  color: #d14;
}
.highlight .sh {
  color: #d14;
}
.highlight .si {
  color: #d14;
}
.highlight .sx {
  color: #d14;
}
.highlight .sr {
  color: #009926;
}
.highlight .s1 {
  color: #d14;
}
.highlight .ss {
  color: #990073;
}
.highlight .s {
  color: #d14;
}
.highlight .na {
  color: #008080;
}
.highlight .bp {
  color: #999999;
}
.highlight .nb {
  color: #0086B3;
}
.highlight .nc {
  color: #445588;
  font-weight: bold;
}
.highlight .no {
  color: #008080;
}
.highlight .nd {
  color: #3c5d5d;
  font-weight: bold;
}
.highlight .ni {
  color: #800080;
}
.highlight .ne {
  color: #990000;
  font-weight: bold;
}
.highlight .nf {
  color: #990000;
  font-weight: bold;
}
.highlight .nl {
  color: #990000;
  font-weight: bold;
}
.highlight .nn {
  color: #555555;
}
.highlight .nt {
  color: #000080;
}
.highlight .vc {
  color: #008080;
}
.highlight .vg {
  color: #008080;
}
.highlight .vi {
  color: #008080;
}
.highlight .nv {
  color: #008080;
}
.highlight .ow {
  color: #000000;
  font-weight: bold;
}
.highlight .o {
  color: #000000;
  font-weight: bold;
}
.highlight .w {
  color: #bbbbbb;
}
.highlight {
  background-color: #f8f8f8;
}
```

</details>

> > Oh yeah, the Heroku previews expire after a certain amount of time. Have you been able to get the site up and running locally? That's the most reliable way to view it. I've also renewed the review app, so it should be accessible again for a while.
> 
> I might be dense here. Currently the local link still goes to Logdown. What directory should I be looking in for the local instance?

https://sass-lang-pr-409.herokuapp.com/blog *should* take you there, but if it was still linking to Logdown something else might be going wrong with the Heroku infrastructure. I'm going to try to land the blog tomorrow so if it's not working this time you'll be able to see it on the live site then.

tatianamac:
> Got it, thanks for educating me on this. Building on the WAI-ARIA example, how would you recommend we handle `tabindex`?

The `tabindex` will remain *structurally* the same (outermost element will receive `tabindex=0` and first innermost child will receive `tabindex=-1` until its focus shifts to its siblings (next child). None of the handling should change, but instead, we're just shifting which element receives the handling (from `li` to `a`), and subsequent role changes. If you're good with that with the context you know about the handling, I can make this change. 🤗

tatianamac:
OK, thanks everyone for your patience here! Below is a summary of where we're at (if you ignore everything above this and review this with latest changes, you should have accurate picture):

✅  `a` 🔗  tags handled
✅  blog reviewed (mostly look good/handled by other global changes already, will do second check once integrated)
✅  Colour contrast stylesheet import sorted

**_Needs Help/Input_**
- **`<aside>`**
I removed the aside from an earlier instance as it was nested within a few layers of div
- **`<dl><dt>`**
@jina , I dug into this a bit more. While the spec you provided is definitely valid for MDN, both Chrome and Axe mark this as a fail for their tests. (If we're using Chrome Lighthouse as the testing standard, this dings us 2 points.
_Recommendation_: Try to recreate layout without nesting divs.
(I tried removing the `divs` from the `components.html.haml` file and it didn't seem to change anything.)
- **tablist**
@Awjin I tried to extract this, but there are quite a few dependencies in Ruby and JS that I don't necessarily want to try to cobble together as I worry I'll break the logic. Two options to solve this:
**Option A** Remove `a` (leaving text inside in the `li` instead) and shift all `a` classes/characteristics up to `li`, check tabfocus to ensure JS logic remains
OR
**Option B** Swap `a` and `li` characteristics such that `a` inherits `role=presentation` and `li` inherits `role=tab`.

Again you all probably have better context than I do for which option is better. From an a11y standpoint, both are sufficient and if behaviour stays the same, will ensure appropriate keyboard focus and no penalties from automatic a11y checkers. Either way, I need someone with more knowledge of Ruby/JS to help, please. 🤗 

Once those last two things are sorted and the blog is up, I will do one final test on all the tools to ensure all systems are clear.

config.rb
```diff
@@ -110,7 +110,7 @@
   content.gsub(%r{^<(h[0-6])(.*?)</\1>}m) do |header_text|
     header = Nokogiri::HTML::DocumentFragment.parse(header_text).children.first
     id = header.attr(:id)
-    header.children.before("<a class='anchor' href='##{id}'></a>") if id
+    header.children.before("<a class='anchor' href='##{id}'><span class='visuallyhidden'>#{header.text} permalink</a>") if id 
     header.to_html
```
jina: The span is not closed. Also (and sorry if this was talked about already and I'm koolaid-manning into this but) would it be better to put the assistive class on the anchor?

source/assets/css/sass.css.scss
```diff
@@ -6,7 +6,8 @@
   'vendor/normalize',
   'vendor/sanitize',
   'vendor/html5-boilerplate',
-  'vendor/syntax',
+  // 'vendor/syntax',
```
jina: we can just remove this

source/assets/css/vendor/_syntax-accessible.scss
```diff
@@ -0,0 +1,265 @@
+// Stylesheet pulled from Rogue stylesheet with colours updated for a11y
+
+.highlight table td { padding: 5px; }
```
jina: can we wrap this in the rem function?

source/assets/css/vendor/_syntax-accessible.scss
```diff
@@ -0,0 +1,265 @@
...
+  .highlight .o {
+    color: #000000;
+    font-weight: bold;
+  }
+  .highlight .w {
+    color: #bbbbbb;
+  }
+  .highlight {
+    background-color: #f8f8f8;
+  }
```
jina: this is super nitpicky but we usually have an extra carriage return at the end of every file.

source/layouts/components/_reference_table_of_contents.haml
```diff
@@ -1,4 +1,4 @@
-%nav.sl-c-list-navigation-wrapper.sl-c-list-navigation-wrapper--collapsible
-  %h3 Contents
+%div.sl-c-list-navigation-wrapper.sl-c-list-navigation-wrapper--collapsible
```
jina: In haml, you don't need to define a `div` — the classes will automatically be a `div` when there is no element defined. :)

source/styleguide/components.html.haml
```diff
@@ -1,5 +1,5 @@
 ---
-title: Components
+dh.title: Components
 ---
```
jina: typo?

config.rb
tatianamac: Ah, thanks! Added the closing tag. (Legit laughed at Koolaidmaning it)

From my understanding, the `a` is what allows the `:hover` that shows the 🔗. So we don't want to visually hide that, just the descriptive text for SRs who would otherwise see an empty link without. Hopefully that makes sense, and open to hearing if I misunderstood the intent/execution of the `a` tag.

jina: ah ok cool

source/assets/css/regions/_banner.scss
```diff
@@ -3,10 +3,13 @@
...
   a {
     display: inline-block;
+  }
+
+  &__brand a {
     border: 0;
   }
 
-  &__brand img {
+  img {
```
Awjin: Are we losing a meaningful inheritance here?

config.rb
Awjin: I think git/hub lost your most recent changes that addressed these resolved comments.

source/assets/css/sass.css.scss
Awjin: Sorry to reopen this convo, but it looks like git/hub lost the changes from [this commit](https://github.com/sass/sass-site/pull/421/commits/be4d0d17ee3888b9747662aaf7739fb0fa37e088).

source/assets/css/vendor/_syntax-accessible.scss
Awjin: Same here.

source/layouts/components/_reference_table_of_contents.haml
Awjin: Same here.

source/styleguide/components.html.haml
Awjin: Same here.

source/assets/css/vendor/_syntax-accessible.scss
Awjin: Same here.

tatianamac:
@jina @nex3  Do you all have any other edits for me? Otherwise I believe this is ready! Thanks for everyone's patience throughout the holidays!

nex3:
I'm happy with it if @Awjin is! @jina?

nex3:
Awesome! Great work Tatiana, thanks so much!

PR #425, opened 2019-12-26
Document configuring through imports

jathak:


source/documentation/at-rules/import.html.md.erb
```diff
@@ -531,6 +531,56 @@ nice API for users of the new module system.
   @forward "reset" as reset-*
 <% end %>
 
+#### Configuring Modules Through Imports
+
+<% impl_status dart: '1.24.0', libsass: false, ruby: false %>
+
+You can configure modules that are loaded through an `@import` simply by
```
nex3: Try to avoid words like "just" and "simply"—they can feel patronizing for some readers.

Also, link "configure modules" to the section describing how default variables work. Users reading this may not have the context to understand exactly what kind of "configuration" you're referring to.

source/documentation/at-rules/import.html.md.erb
```diff
@@ -531,6 +531,56 @@ nice API for users of the new module system.
...
+defining global variables prior the `@import` that first loads that module.
+
+Note that changes to the configuration after the module is first loaded will be
+ignored.
```
nex3: I find this kind of confusing. Either elaborate on it by describing exactly the situation in which this might be a problem (possibly within a `<% heads_up do %>` block), or leave it out if you don't think it's likely to be a substantial problem in practice.

source/documentation/at-rules/import.html.md.erb
```diff
@@ -531,6 +531,56 @@ nice API for users of the new module system.
...
+  }
+  ---
+  // _library.import.scss
+  @forward 'library' as lib-*;
+  ---
+  // style.sass
+  $lib-color: green;
+  @import "library";
+  $lib-color: red;
+  @import "library";
```
nex3: I think this is just going to produce more net confusion. Examples on the doc site don't have to be 100% comprehensive (that's what the spec is for), so it's more important that they show the most focused version of a feature possible.

source/documentation/at-rules/import.html.md.erb
```diff
@@ -569,16 +566,16 @@ ignored.
...
   }
+<% end %>
 
-  a {
-    color: green;
-  }
+<% heads_up do %>
+Modules are only loaded once, so if you change the configuration after you
+`@import` a module for the first time (even indirectly), the change will be
+ignored if you `@import` the module again.
```
nex3: Nit: intend this two spaces.

Issue #418, opened 2019-12-03
Document new math functions


nex3:
See https://github.com/sass/sass/issues/851.

PR #427, opened 2020-01-16
Documentation for new built-in math functions.

Awjin:
Closes https://github.com/sass/sass-site/issues/418


source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
+---
+title: sass:math
+---
+
+<%= partial '../snippets/built-in-module-status' %>
+
+## Variables
+
+<% function 'math.$e' do %>
+  Equal to the value of the mathematical constant `e`.
```
nex3: It's probably a good idea to link to the Wikipedia pages for most of these new members. Functions like `clamp()` doesn't need that since you can easily describe its behavior in prose, but it's good practice to assume that readers won't necessarily be coming from a mathematical background.

nex3: All the new members here should have their own `impl_status` tags indicating when they were added to Dart Sass.

nex3:
```suggestion
  Equal to the value of the mathematical constant *e*.
```

Since you're referring to a mathematical constant here rather than programming language syntax, it's probably a good idea to use a different typographic convention. Italics are often used when math is included in prose.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  <% example(autogen_css: false) do %>
+    @debug math.$e; // 2.7182818285
+    ===
+    @debug math.$e // 2.7182818285
+  <% end %>
+<% end %>
+
+
+<% function 'math.$pi' do %>
+  Equal to the value of the mathematical constant `pi`.
```
nex3:
```suggestion
  Equal to the value of the mathematical constant *π*.
```

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+
+    $widths: 50px, 30px, 100px
+    @debug math.min($widths...)  // 30px
+  <% end %>
+<% end %>
+
+
+<% function 'math.round($number)',
+            'round($number)',
+            returns: 'number' do %>
```
nex3: Nit: This signature only needs to be line-wrapped if it's particularly long, which it doesn't look like it is.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+    ===
+    @debug math.ceil(4)  // 4
+    @debug math.ceil(4.2)  // 5
+    @debug math.ceil(4.9)  // 5
+  <% end %>
+<% end %>
+
+
+<% function 'math.clamp($min, $number, $max)', returns: 'number' do %>
+  Clamps `$number` in between `$min` and `$max`.
```
nex3: Be more explicit about this. If it were clear to a reader what "clamps" meant, they probably wouldn't need to read the documentation :smile:.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  Returns the length of the *n*-dimensional vector that has components equal to
+  each `$number`.
+
+  The numbers must either all have compatible units, or all be unitless. And
+  since the numbers' units may differ, the output takes the unit of the first
+  number.
+
+  A list of numbers can be passed [using `...`][].
+
+  [using `...`]: ../at-rules/function#passing-arbitrary-arguments
```
nex3: I don't know if this is worth calling out explicitly, especially since it's covered in the example below.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+    ===
+    @debug math.abs(10px) // 10px
+    @debug math.abs(-10px) // 10px
+  <% end %>
+<% end %>
+
+
+<% function 'math.hypot($number...)', returns: 'number' do %>
+  Returns the length of the *n*-dimensional vector that has components equal to
+  each `$number`.
```
nex3: Vector lengths are another thing that are likely to puzzle non-mathy users, even with a Wikipedia entry. I might add a short abstract example here, like:

> For example, for three numbers *a*, *b*, and *c*, this returns the square root of *a² + b² + c²*.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  <% end %>
+<% end %>
+
+
+## Exponential Functions
+
+<% function 'math.log($number, $base: null)',
+            returns: 'number' do %>
+  Returns the logarithm of `$number` with respect to `$base`. If `$base` is
+  [`null`][], the [natural log][] is calculated.
```
nex3: I think linking `null` here is overkill; we don't generally link it elsewhere.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  <% end %>
+<% end %>
+
+
+## Trigonometric Functions
+
+<% function 'math.cos($number)', returns: 'number' do %>
+  Returns the [cosine][] of `$number`.
+
+  `$number` must be an angle (its units are compatible with `deg`) or unitless.
```
nex3:
```suggestion
  `$number` must be an angle (its units must be compatible with `deg`) or unitless.
```

Also below.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  <% example(autogen_css: false) do %>
+    @debug math.atan(10); // 84.2894068625deg
+    ===
+    @debug math.atan(10) // 84.2894068625deg
+  <% end %>
+<% end %>
+
+
+<% function 'math.atan2($y, $x)', returns: 'number' do %>
+  Returns the [2-argument arctangent] of `$y` and `$x` in `deg`.
```
nex3: Style nit:

```suggestion
  Returns the [2-argument arctangent][] of `$y` and `$x` in `deg`.
```

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,494 @@
...
+  `$y` and `$x` must have compatible units or be unitless.
+
+  [2-argument arctangent]: https://en.wikipedia.org/wiki/Atan2
+
+  <% fun_fact do %>
+    `atan2($y, $x)` is distinct from `atan($y / $x)` because it preserves the
+    quadrant of the point in question. For example, `atan2(1, -1)` corresponds
+    to the point `(-1, 1)` and returns `135deg`. In contrast, `atan(1 / -1)` and
+    `atan(-1 / 1)` resolve first to `atan(-1)`, so both return `-45deg`.
+  <% end %>
```
nex3: Nice :smile:

source/documentation/modules/math.html.md.erb
Awjin: Done.

Awjin: Done.

source/documentation/modules/math.html.md.erb
Awjin: Done.

source/documentation/modules/math.html.md.erb
Awjin: Done.

source/documentation/modules/math.html.md.erb
Awjin: So you're telling me

> clamp /klamp/ *verb*:
> to clamp

isn't good enough? 😛 

Fixed.

source/documentation/modules/math.html.md.erb
Awjin: Removed here and elsewhere.

source/documentation/modules/math.html.md.erb
Awjin: That's a nice explainer, thanks. Added.

source/documentation/modules/math.html.md.erb
Awjin: Done.

source/documentation/modules/math.html.md.erb
Awjin: Done.

source/documentation/modules/math.html.md.erb
Awjin: Done.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+    ===
+    @debug math.$e // 2.7182818285
+  <% end %>
+<% end %>
+
+
+<% function 'math.$pi' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Equal to the value of the [mathematical constant *π*].
```
nex3: Nit:

```suggestion
  Equal to the value of the [mathematical constant *π*][].
```

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+  <% end %>
+<% end %>
+
+
+<% function 'math.clamp($min, $number, $max)', returns: 'number' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Restricts `$number` to the range between `$min` and `$max`. If `$number`
+  crosses a boundary (`$min` or `$max`), this returns the value of that
+  boundary.
```
nex3: I'd be even more explicit:

> If `$number` is less than `$min` this returns `$min`, and if it's greater than `$max` this returns `$max`.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+  <% end %>
+<% end %>
+
+
+## Exponential Functions
+
+<% function 'math.log($number, $base: null)', returns: 'number' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Returns the logarithm of `$number` with respect to `$base`. If `$base` is
```
nex3: I'd like "logarithm" as well.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+    @debug math.abs(10px) // 10px
+    @debug math.abs(-10px) // 10px
+  <% end %>
+<% end %>
+
+
+<% function 'math.hypot($number...)', returns: 'number' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Returns the length of the *n*-dimensional vector that has components equal to
```
nex3: I'd link "vector".

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+    @debug math.log(10) // 2.302585093
+    @debug math.log(10, 10) // 1
+  <% end %>
+<% end %>
+
+
+<% function 'math.pow($base, $exponent)', returns: 'number' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Raises `$base` to the power of `$exponent`.
```
nex3: I'd link "to the power of" to https://en.wikipedia.org/wiki/Exponentiation.

source/documentation/modules/math.html.md.erb
```diff
@@ -0,0 +1,511 @@
...
+    @debug math.pow(100, 1/3) // 4.6415888336
+    @debug math.pow(5, -2) // 0.04
+  <% end %>
+<% end %>
+
+
+<% function 'math.sqrt($number)', returns: 'number' do %>
+  <% impl_status dart: '1.25.0', libsass: false, ruby: false %>
+
+  Returns the square root of `$number`.
```
nex3: I'd like "square root".

PR #428, opened 2020-01-16
Increase the indent size for function definition docs.

Awjin:
While updating the math functions documentation, I realized that the large amount of text is making it hard to navigate the docs.

Increasing the indent size seems to improve readability. Attached are before/after pics to show the difference.

docs-before

docs-after

### DIFF
--- a/source/assets/css/components/_callouts.scss
+++ b/source/assets/css/components/_callouts.scss
@@ -15,12 +15,15 @@
   }
 
   &--function {
-  padding: 0 2.25rem .25rem;
+    padding-bottom: .25rem;
+    padding-top: 0;
 
     .signature {
       // Make sure permalinks are visible.
       overflow: visible;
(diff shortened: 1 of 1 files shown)


source/assets/css/components/_callouts.scss
```diff
@@ -15,19 +15,12 @@
...
 
     .signature {
       // Make sure permalinks are visible.
       overflow: visible;
-
-      margin: {
-        right: -1rem;
-        left: -1rem;
-      };
+      margin-left: -2.25rem;
```
jina: can we make these changes in the larger breakpoint only? right now in production its even both sides (margins & paddings) but this change (meant to adjust for larger size) will look wrong on smaller screen sizes (like a mobile web size).

Awjin: Thanks, fixed.

PR #449, opened 2020-05-15
docs: Fix mixin example Sass error for top level &

szTheory:
SassError: Top-level selectors may not contain the parent selector "&".


Awjin:
Thanks for the contribution! However, I think this makes the example's intent unclear. The example should be focused on illustrating content blocks. And in general, creating a fallback like this is an implementation detail that's dependent on the context and usage of the mixin, which we try not to prescribe in the documentation.

szTheory:
The only problem is the example gives a compiler error if you copy and paste it as is. But I understand.

PR #455, opened 2020-06-02
Fix behavioral differences link

kuzivany:
Fixes the _intentional behavior differences_ link fragment so it points to the correct section


jathak:
Thanks!

PR #463, opened 2020-08-04
Remove Chocolatey install instructions

jathak:
The Chocolatey release has been out of date for months, so we shouldn't be recommending it until we're able to reliably keep it up to date.


nex3:
Can you file an issue to re-enable this once chocolatey support is fixed? Also maybe an issue on cli-pkg to fix Chocolatey support?

jathak:
Done. See #464 and google/dart_cli_pkg#49.

PR #467, opened 2020-08-13
Add Linux brew install instruction.

Awjin:


nex3:
Thanks for thinking of this!

PR #468, opened 2020-08-14
Removed unnecessary line separator character

martingreilinger:
**Description**

The line separator '\u2028' character was removed (looked out of place).

Highlighted the removed _character_ in the screenshot below for reference:
removed-character

### DIFF
--- a/source/styleguide/index.html.md
+++ b/source/styleguide/index.html.md
@@ -7,5 +7,5 @@ introduction: >
 
 ### Site Objectives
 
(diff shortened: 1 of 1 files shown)


jina:
I'm a little confused at this pull request. The diff does not show any actual change. And when I view the page, there is no extra space. It doesn't exist?

jina:
ahhh. chrome shows me nothing but safari shows it. strange bug!

PR #479, opened 2020-09-28
Fix option name for sourceMapEmbed in example

### LINKED ISSUE #478: sourceMapEmbed or embedSourceMap?
The JavaScript API docs show conflicting information about the option for embedding a source map in the compiled CSS:

https://github.com/sass/sass-site/blob/165dc8f293028c75c1920318b718a0a9d091ecc8/source/documentation/js-api.html.md.erb#L593-L605

My understanding is that the example code is incorrect and should be changed to `sourceMapEmbed`. If this is the case I'm happy to open a pull request to resolve this issue.

Thanks in advance!

greatislander:
Resolves #478.

### DIFF
--- a/source/documentation/js-api.html.md.erb
+++ b/source/documentation/js-api.html.md.erb
@@ -600,7 +600,7 @@ CSS.
 sass.renderSync({
   file: "style.scss",
   sourceMap: "out.map",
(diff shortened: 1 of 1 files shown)


Awjin:
Thanks for the fix!

Issue #480, opened 2020-10-03
Sass docs recommends `@use` over `@import` without clarifying node-sass does not support `@use`


wu-lee:
The Sass docs say [here][1]:

> The Sass team discourages the continued use of the @import rule. Sass will gradually phase it out over the next few years, and eventually remove it from the language entirely. Prefer the @use rule instead.

However, I cannot get `@use` to work. I now know the reason, which is that node-sass does not support it. However, there are no errors raised at the point of the `@use` declaration. Instead they suggest an error later in the source where I attempt to utilise the imported symbols:

> Invalid CSS after "...nd-color: color": expected expression (e.g. 1px, bold), was ".adjust($primary-co"

Only after an intensive search did I find out the real cause, from posts such as this one, which also illustrates the case in point:

https://stackoverflow.com/questions/60920833/gatsby-fails-after-using-sass-files-with-use-sasscolor

To help future users avoid an unnecessary hunt, at least until node-sass is no-longer the default in ecosystems like Gatsby, please could the documentation state clearly that node-sass does not support `@use`, next to the advice to prefer `@use` to over `@import`

[1]: https://sass-lang.com/documentation/at-rules/import

nex3:
At the very top of the `@use` documentation is a compatibility header that describes exactly which implementation support it, as well as a drop-down that provides more details:

![image](https://user-images.githubusercontent.com/188/95142625-f676d000-0728-11eb-9fe4-eeeae9038d53.png)

wu-lee:
I don't deny this. I humbly suggest that something more along the lines of the very noticeable warnings like this (image below) are warrented. Especially so, because the error messages from libsass are misleading:

![Screenshot_20201006_103029](https://user-images.githubusercontent.com/67825/95184354-24a6ef00-07bf-11eb-8d6b-93b115a19cbe.png)

Possibly this seems a small thing to you, but for others it could save a lot of time.

Awjin:
@wu-lee This was fixed in https://github.com/sass/sass-site/pull/481

PR #481, opened 2020-10-05
Add explainer that @use is only availabe in Dart Sass

Awjin:
Closes https://github.com/sass/sass-site/issues/480

### DIFF
--- a/source/documentation/at-rules/import.html.md.erb
+++ b/source/documentation/at-rules/import.html.md.erb
@@ -23,7 +23,7 @@ required to have quotes.
   [gradually phase it out][] over the next few years, and eventually remove it
   from the language entirely. Prefer the [`@use` rule][] instead. (Note that
   only Dart Sass currently supports `@use`. Users of other implementations must
-  use the @import rule instead.)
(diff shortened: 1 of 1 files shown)


source/documentation/at-rules/import.html.md.erb
```diff
@@ -21,7 +21,9 @@ required to have quotes.
 <% heads_up do %>
   The Sass team discourages the continued use of the `@import` rule. Sass will
   [gradually phase it out][] over the next few years, and eventually remove it
-  from the language entirely. Prefer the [`@use` rule][] instead.
+  from the language entirely. Prefer the [`@use` rule][] instead. (Note that
+  only Dart Sass currently supports `@use`. Users of other implementations must
+  use the @import rule instead.)
```
nex3, line 26:
```suggestion
  use the `@import` rule instead.)
```

nex3 (approved):

source/documentation/at-rules/import.html.md.erb
Awjin: Thanks

Issue #476, opened 2020-09-17
Documented nested map support in sass:map functions


nex3:
See sass/sass#1739 and sass/sass#2836.

PR #484, opened 2020-10-06
Nested maps documentations

Awjin:
Closes https://github.com/sass/sass-site/issues/476

### DIFF
--- a/source/documentation/modules/map.html.md.erb
+++ b/source/documentation/modules/map.html.md.erb
@@ -24,10 +24,8 @@ title: sass:map
 
 
 <% function 'map.deep-merge($map1, $map2)', returns: 'map' do %>
-  Identical to [`map.merge()`][], except that nested map values are *also*
+  Identical to [`map.merge()`](#merge), except that nested map values are *also*
   recursively merged.
-
-  [`map.merge()`][#merge]
 
   <% example(autogen_css: false) do %>
     $helvetica-light: (
@@ -100,6 +98,8 @@ title: sass:map
     // ("medium": 500, "bold": 700)
   <% end %>
 
+  ---
+
   If `$keys` is not empty, the map targeted for updating is nested within
   `$map`. From left to right, including `$key` and excluding the last key in
   `$keys`, these keys form a path of nested maps that leads to the targeted map.
@@ -151,6 +151,8 @@ title: sass:map
 
   <%= partial 'code-snippets/example-map-get' %>
 
+  ---
+
   If `$keys` is not empty, the map targeted for searching is nested within
   `$map`. From left to right, including `$key` and excluding the last key in
   `$keys`, these keys form a path of nested maps that leads to the targeted map.
@@ -158,7 +160,8 @@ title: sass:map
   Returns the value in the targeted map associated with the last key in `$keys`.
 
   Returns [`null`][] if the map does not have a value associated with the key,
-  or if any key in `$keys` references a value that is not a map.
+  or if any key in `$keys` is missing from a map or references a value that is
+  not a map.
 
   <% example(autogen_css: false) do %>
     $fonts: (
@@ -202,6 +205,8 @@ title: sass:map
     @debug map.has-key($font-weights, "bolder") // false
   <% end %>
 
+  ---
+
   If `$keys` is not empty, the map targeted for searching is nested within
   `$map`. From left to right, including `$key` and excluding the last key in
   `$keys`, these keys form a path of nested maps that leads to the targeted map.
@@ -209,8 +214,8 @@ title: sass:map
   Returns true if the targeted map contains a value associated with the last key
   in `$keys`.
 
-  Returns false if it does not, or if any key in `$keys` references a value
-  that is not a map.
+  Returns false if it does not, or if any key in `$keys` is missing from a map
+  or references a value that is not a map.
 
   <% example(autogen_css: false) do %>
     $fonts: (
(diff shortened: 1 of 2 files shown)


source/documentation/modules/map.html.md.erb
```diff
@@ -49,8 +116,53 @@ title: sass:map
 <% end %>
 
 
+<% function 'map.set($map, $key, $value)',
+            'map.set($map, $keys..., $key, $value)',
```
Awjin, line 120: @nex3 Describing the signature like this (vs `map.set($map, $args...)`) massively simplifies the explanation. Is it worth reflecting this in the spec?

nex3: The trouble is that this isn't valid Sass arglist declaration syntax (although arguably it should be). In order to use it in the spec, we'd need to formally specify exactly what it means and how it translates into actual behavior.

Even using it here worries me a little because it implies that `$key` and `$value` can be passed by name in the second signature when in fact they can't. It may still be worth it with a note explaining that, though.

Awjin: Oof, right, I forgot that arglists always have to follow positional args. Maybe I'll change this signature back to `$args`, but then "deconstruct" it as `$map, $keys..., $key, $value` in the explanation

source/documentation/modules/map.html.md.erb
```diff
@@ -28,78 +67,297 @@ title: sass:map
...
     $font-weights: ("regular": 400, "medium": 500, "bold": 700)
 
-    @debug map.has-key($font-weights, "regular")  // true
-    @debug map.has-key($font-weights, "bolder")  // false
+    @debug map.has-key($font-weights, "regular") // true
+    @debug map.has-key($font-weights, "bolder") // false
+  <% end %>
+
+  If `$keys` is not empty, the map targeted for searching is nested within
+  `$map`. From left to right, including `$key` and excluding the last key in
```
nex3, line 75: Avoid passive voice here. Maybe instead: "If `$keys` is not empty, follows them from left to right to find a nested map to check." Using "follows" here mirrors your explanation of nested maps above. Same suggestion for other functions.

source/documentation/modules/map.html.md.erb
```diff
@@ -28,78 +67,297 @@ title: sass:map
...
+  `$keys`, these keys form a path of nested maps that leads to the targeted map.
+
+  Returns true if the targeted map contains a value associated with the last key
+  in `$keys`.
+
+  Returns false if it does not, or if any key in `$keys` references a value
+  that is not a map.
```
nex3, line 82:
```suggestion
  Returns false if it does not, or if any key in `$keys` is missing from a
  map or references a value that is not a map.
```

source/documentation/modules/map.html.md.erb
```diff
@@ -28,78 +67,297 @@ title: sass:map
...
+  [`null`]: ../values/null
+<% end %>
+
+
+<% function 'map.set($map, $key, $value)',
+            'map.set($map, $keys..., $key, $value)',
+            returns: 'map' do %>
+  <% heads_up do %>
+    In practice, the actual arguments to `map.set($map, $keys..., $key, $value)`
+    are passed as `map.set($map, $args). They are described here as `$map,
```
nex3, line 152:
```suggestion
    are passed as `map.set($map, $args...). They are described here as `$map,
```

This is relevant because you can't pass `$args` by name.

source/documentation/modules/map.html.md.erb
```diff
@@ -28,78 +67,297 @@ title: sass:map
     ===
     $font-weights: ("regular": 400, "medium": 500, "bold": 700)
 
-    @debug map.has-key($font-weights, "regular")  // true
-    @debug map.has-key($font-weights, "bolder")  // false
+    @debug map.has-key($font-weights, "regular") // true
+    @debug map.has-key($font-weights, "bolder") // false
+  <% end %>
+
```
nex3, line 73: We usually separate different overloads of the same function with `---`, like you did for `map.set()`.

source/documentation/modules/map.html.md.erb
```diff
@@ -28,78 +67,297 @@ title: sass:map
...
 
-    // map.merge() can be used to add a single key/value pair to a map.
-    @debug map.merge($light-weights, ("lighter": 200))
-    // ("lightest": 100, "light": 300, "lighter": 200)
+<% function 'map.deep-merge($map1, $map2)', returns: 'map' do %>
+  Identical to [`map.merge()`][], except that nested map values are *also*
+  recursively merged.
+
+  [`map.merge()`][#merge]
```
nex3, line 309:
```suggestion
  [`map.merge()`]: #merge
```

or just include this inline above like ```[`map.merge()`](#merge)```.

nex3 (changes requested):

source/documentation/modules/map.html.md.erb
```diff
@@ -4,21 +4,60 @@ title: sass:map
...
 
-  [`null`]: ../values/null
+  <% example(autogen_css: false) do %>
+    $config: (a: (b: (c: d)));
+    @debug map.get($config, a, b, c)); // d
+    ===
+    $config: (a: (b: (c: d)))
+    @debug map.get($config, a, b, c)) // d
+  <% end %>
+<% end %>
```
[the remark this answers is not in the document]
Awjin: Thanks!

source/documentation/modules/map.html.md.erb
```diff
@@ -4,21 +4,60 @@ title: sass:map
...
 
-  <%= partial 'code-snippets/example-map-get' %>
+
+<% function 'map.keys($map)', 'map-keys($map)', returns: 'list' do %>
```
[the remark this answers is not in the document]
Awjin: I'll resort them alphabetically

nex3 (approved):

PR #485, opened 2020-10-07
Elixir sass

scottdavis:


source/libsass.html.haml
```diff
@@ -178,6 +183,7 @@ introduction: >
       - [SassC](#sassc)
       - [Crystal](#crystal)
       - [Go](#go)
+      - [Elixir](#elixir)
```
Awjin, line 186: This should be one line above Go.

CONTRIBUTING.md
```diff
@@ -31,6 +31,8 @@ bundle exec rake sass:import
 bundle exec middleman
 ```
 
+Try this [gist](https://gist.github.com/fernandoaleman/868b64cd60ab2d51ab24e7bf384da1ca) if you are having v8 trouble on macosx Catalina
```
Awjin, line 34: Since this isn't a problem all contributors will run into, IMO we should leave it out of the contributing doc.

Awjin (changes requested):
Thanks for contributing! Had a few minor comments.

PR #486, opened 2020-10-07
Resolved typo in argument lists example

martingreilinger:
Resolved the typo inside this "argument lists" example.
(in both the `SASS` and `SCSS` snippets)

Highlighted the changed character in the screenshot below:
Before | After
----------|--------
before|after


Awjin:
Good catch, thanks!

PR #488, opened 2020-10-08
Correct and reformat sample codes

ngdangtu-vn:
This PR is about: (only content is changed)
- Update a sass sample code to match with scss sample code
- Correct sass sample codes' syntax
- Reformat comment in 1 sass sample code (for readability)
- Reformat 1 scss sample code (for matching coding style)

This PR should not cause any serious problem for the website and its content. Even without correcting it, readers should be understand the document. But, leaving it like this is not a good way.

### DIFF
--- a/source/documentation/at-rules/use.html.md.erb
+++ b/source/documentation/at-rules/use.html.md.erb
@@ -364,11 +364,13 @@ another mixin to inject your styles.
 
   @include styles
     code
-      border-radius: $border-radius
-      box-shadow: $box-shadow
+      border-radius: $-border-radius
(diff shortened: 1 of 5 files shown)


source/documentation/at-rules/use.html.md.erb
```diff
@@ -364,11 +364,13 @@ another mixin to inject your styles.
 
   @include styles
```
Awjin, line 365:
```suggestion
  @mixin styles
```

Awjin (commented):
Thank you for these fixes!

Awjin (approved):

PR #490, opened 2020-10-08
Add missing backtick at the end of inline code

chriskirknielsen:
Caught this on mobile as layout was broken.


Awjin:
Thanks for the fix!

PR #496, opened 2020-10-21
Spacing improvement

danielhaim1:
- Added a spacer to the follow button on footer (mobile only)
- Added missing spacer from page titles (mobile + desktop)

### Nav (before/after)
image

### Footer (before/after)
image

### DIFF
--- a/source/assets/css/regions/_banner.scss
+++ b/source/assets/css/regions/_banner.scss
@@ -1,6 +1,10 @@
 .sl-r-banner {
   position: relative;
 
+  // add bottom padding to increase space
+  // above page titles (visible on mobile site)
+  padding-bottom: 3rem;
+
   a {
     display: inline-block;
   }
--- a/source/assets/css/components/_lists.scss
+++ b/source/assets/css/components/_lists.scss
@@ -10,6 +10,12 @@
 
   &-horizontal-wrapper {
 
+    @include sl-breakpoint--medium {
+      // add margin-bottom space
+      // to the twitter follow button
+      margin-bottom: 1rem;
+    }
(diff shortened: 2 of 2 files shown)


jina:
Can I see a screenshot of these on a larger screen?

(@nex3 for some reason this didn't deploy so I could preview)

nex3:
I believe it only deploys for repo collaborators, so that people can't use the preview to just host whatever site on our Heroku instance.

jina:
gotcha. makes sense! @danielhaim1 so yeah the screenshots of how this looks on a larger screen would be super helpful. thanks!

source/assets/css/regions/_banner.scss
```diff
@@ -1,6 +1,10 @@
 .sl-r-banner {
   position: relative;
 
+  // add bottom padding to increase space
+  // above page titles (visible on mobile site)
+  padding-bottom: 3rem;
+
```
jina, line 7: this was the reason I wanted to see a screenshot. It's not scoped to a media query so it would be adding space to the desktop web view as well. I think if we can just scope this, too, we're all set.

jina (changes requested):

source/assets/css/regions/_banner.scss
danielhaim1: @jina here's a screenshot, the improvement is primarily targeted for mobile, but also adds some spacing for desktop as well. I plan to go back into this area and improve it further in a couple weeks.

Before/After:
image

danielhaim1:
done 👍 @nex3 @jina

PR #493, opened 2020-10-16
Add documentation for new HWB functions

nex3:
See sass/sass#2834
Closes #477

PR #508, opened 2020-11-24
Update the documentation about the `hwb` function

johannpinson:
After looking for the changelog of dart-sass, I discover the new `hwb()` function added recently.
But in the documentation, the function is mentionned without the `color.` prefix to respect Sass module system.

After a test, this is the result:
```scss
@debug hwb(210, 0%, 60%); // DEBUG: hwb(210, 0%, 60%)
@debug color.hwb(210, 0%, 60%); // DEBUG: #003366
```

This PR updates the docs following the example of the `color.whiteness()` function, also added from the PR #493 by @nex3

### DIFF
--- a/source/documentation/modules/color.html.md.erb
+++ b/source/documentation/modules/color.html.md.erb
@@ -436,9 +436,9 @@ SIGNATURE
 <% end %>
 
 
-<% function 'hwb($hue $whiteness $blackness)',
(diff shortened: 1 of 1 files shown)


jathak:
Thanks for this!

johannpinson:
Good catch, I didn't notice I forgot one line 😉

Thanks for the merge.

PR #529, opened 2021-03-18
Revert word-break to normal for impl-status titles

xmo-odoo:
If the "compatibility" label and browser width are just the wrong one, "break-word" can lead to browsers deciding to line-wrap every "ss" but *just the "ss" from "Dart Sass", "LibSass" and "Ruby Sass", which looks very strange and is quite hard to read.

I've a relatively large screen (34" wide), so I lay out the screen in thirds (so about 1146 wide per window). Using Firefox (FDE 87.0b9), this is apparently exactly the worst possible width for http://bit.ly/ExtendCompound as every single "ss" gets wrapped to the next line:
Screenshot_2021-03-18 Sass extend
After:
Screenshot_2021-03-18 Sass extend(1)
Disclaimer: I didn't try to build the site, I checked which change should work using the web inspector, then tried to find out where the properties were being defined, then performed the change via github's web UI.

### DIFF
--- a/source/assets/css/components/_lists.scss
+++ b/source/assets/css/components/_lists.scss
@@ -126,6 +126,8 @@
   &.impl-status {
     font-size: 0.8em;
     margin-top: -1em;
+    
+    dt { word-break: normal; }
 
     a {
       transition: transform 0.1s;


Awjin:
Thanks!

Issue #531, opened 2021-04-16
[Content] Why CSS has not a feature of variables?


MenaiAla:
In the definition of processor, you mentioned that CSS has not a feature of **variables** while CSS has this feature using the function `var()`.

```
CSS on its own can be fun, but stylesheets are getting larger, more complex, and harder to maintain. 

This is where a preprocessor can help. Sass lets you use features that don't exist in CSS yet like variables, 

nesting, mixins, inheritance, and other nifty goodies that make writing CSS fun again.
```

PR #533, opened 2021-04-19
Update intro.

Awjin:
WDYT of this addition?

Fixes https://github.com/sass/sass-site/issues/531

### DIFF
--- a/source/guide.html.haml
+++ b/source/guide.html.haml
@@ -27,8 +27,8 @@ introduction: >
 
     CSS on its own can be fun, but stylesheets are getting larger, more
     complex, and harder to maintain. This is where a preprocessor can help.
-    Sass lets you use features that don't exist in CSS yet like nesting,
-    mixins, inheritance, and other nifty goodies that help you write robust,
+    Sass has features that don't exist in CSS yet like nesting, mixins,
+    inheritance, and other nifty goodies that help you write robust,
     maintainable CSS.
 
     Once you start tinkering with Sass, it will take your preprocessed Sass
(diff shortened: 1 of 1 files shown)


nex3:
I like the idea of replacing "variables" with a more modern CSS feature, but I worry that "modules" without any other context might lead people to think that Sass supports something like [CSS Modules](https://github.com/css-modules/css-modules). Maybe "style re-use"? Or is that too abstract?

Awjin:
It might be too abstract compared to the other goodies listed. Hmm. How about "scoped imports"?

nex3:
Feels weird to call that out since it's something we're actively deprecating. Maybe just "custom functions"?

jina:
can just remove the word instead of replacing it.

source/guide.html.haml
```diff
@@ -27,7 +27,7 @@ introduction: >
 
     CSS on its own can be fun, but stylesheets are getting larger, more
     complex, and harder to maintain. This is where a preprocessor can help.
-    Sass lets you use features that don't exist in CSS yet like variables,
+    Sass lets you use features that don't exist in CSS yet like modules,
     nesting, mixins, inheritance and other nifty goodies that make writing CSS
     fun again.
```
jina, line 32: While we are in here, can we also change the last phrase? "make [noun] [adjective] again" has been ruined for me by a certain public figure.

jina (changes requested):

source/guide.html.haml
jina: I also very much dislike the notion that CSS isn't fun. I love writing CSS.

Awjin: I think the original intent was that maintaining thousands lines of pure CSS isn't necessarily fun, but I get what you're saying. Made some edits.

jina: thank you! ❤️

jina (approved):

PR #534, opened 2021-04-25
Add documentation of -w shorthand for --watch flag

hendotcat:
This is the follow-up documentation change for https://github.com/sass/dart-sass/pull/1276. The diff's a little noisy because of the paragraph reflowing to fit the width limit, but the only new thing in there is the `(abbreviated '-w')` part.


Awjin:
Thanks for thinking of this!

PR #539, opened 2021-05-12
Publicize the slash-div deprecation page

nex3:
See sass/sass#2565


jathak (approved):

stof:
There is one issue here: https://sass-lang.com/documentation/modules/math does not document `math.div` while this new doc page asks to use it.

nex3:
Good catch! I'll have to add docs for that and `list.slash()`.

PR #544, opened 2021-05-20
Update all existing documentation for math.div() and list.slash()

### LINKED ISSUE #543: add breaking change '⚠️ Heads up!' callout to 'Slash-Separated Values' section under numeric operators
Under [/documentation/operators/numeric#slash-separated-values](https://sass-lang.com/documentation/operators/numeric#slash-separated-values) the conditions are specified for which Sass will use `/` as a dividing numeric operator:

```
Sass will print the result as slash-separated instead of divided unless one of these conditions is met:

   * Either expression is anything other than a literal number.
   * The result is stored in a variable or returned by a function.
   * The operation is surrounded by parentheses, unless those parentheses are outside a list that contains the operation.
   * The result is used as part of another operation (other than /).
```

It could be helpful to have a '⚠️ Heads up!' callout there referencing [Breaking Change: Slash as Division](https://sass-lang.com/documentation/breaking-changes/slash-div).

nex3:
Closes #543
See #539

### DIFF
--- a/source/documentation/breaking-changes/slash-div.html.md.erb
+++ b/source/documentation/breaking-changes/slash-div.html.md.erb
@@ -7,7 +7,7 @@ introduction: >
   `/` as a separator.
 ---
 
-<% impl_status dart: false, libsass: false, ruby: false %>
+<% impl_status dart: 'partial', libsass: false, ruby: false %>
 
 Today, Sass uses [complex heuristics][] to figure out whether a `/` should be
 treated as division or a separator. Even then, as a separator it just produces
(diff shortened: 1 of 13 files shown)


jathak (approved):

source/documentation/breaking-changes/slash-div.html.md.erb
```diff
@@ -7,7 +7,7 @@ introduction: >
   `/` as a separator.
 ---
 
-<% impl_status dart: false, libsass: false, ruby: false %>
+<% impl_status dart: 'partial', libsass: false, ruby: false %>
```
stof, line 10: This looks quite weird on the website, as string values are expected to be versions: 
image

nex3: Good catch, this is supposed to be `:partial` (which `impl_status` has special handling for).

PR #546, opened 2021-05-24
Add docs for reassigning module vars.

Awjin:


source/documentation/at-rules/use.html.md.erb
```diff
@@ -378,6 +378,39 @@ another mixin to inject your styles.
...
+  library.$color: blue
+  ---
+  // style.sass
+  @use 'library'
+  @use 'override'
+  @debug library.$color  //=> blue
+<% end %>
+
+Note that built-in module variables (e.g. `math.$pi` from the math module)
+cannot be reassigned.
```
nex3, line 412: Might be worth making this a `heads_up`.

nex3 (commented):
Since this originally came up because people were surprised you could reassign variables specifically with `@use ... as *`, it might be worth explicitly calling that out.

stof:
I agree about the need to mention the case of `as *` too.

PR #550, opened 2021-05-26
Use a more modern mixin example in the tutorial

### LINKED ISSUE #473: Improve mixin and extend docs in https://sass-lang.com/guide

nex3:
Closes #473

### DIFF
--- a/source/guide.html.haml
+++ b/source/guide.html.haml
@@ -239,19 +239,36 @@ introduction: >
 
   - example do
     :plain
-      @mixin transform($property) {
-        -webkit-transform: $property;
-        -ms-transform: $property;
-        transform: $property;
+      @mixin theme($theme: DarkGray) {
+        background: $theme;
+        box-shadow: 0 0 1px rgba($theme, .25);
+        color: #fff;
       }
-      .box { @include transform(rotate(30deg)); }
+
+      .info {
+        @include theme;
+      }
+      .alert {
+        @include theme($theme: DarkRed);
+      }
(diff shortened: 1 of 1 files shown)


source/guide.html.haml
```diff
@@ -239,19 +239,36 @@ introduction: >
...
       }
-      .box { @include transform(rotate(30deg)); }
       ===
-      =transform($property)
-        -webkit-transform: $property
-        -ms-transform: $property
-        transform: $property
-      .box
-        +transform(rotate(30deg))
+      @mixin theme($theme: DarkGray)
```
jina, line 258: are we discouraging the shorthand for the original syntax? (`=` & `+`)?

jina: ok no worries! I just know the previous example used them. but it's not necessary. :)

jina (approved):

PR #551, opened 2021-05-26
Write a blog post about the discontinuation of node-fibers

nex3:


source/blog/029-node-fibers-discontinued.md
```diff
@@ -0,0 +1,160 @@
...
+to help an engineer who's willing to give it a shot.
+
+This isn't a contribution for the faint of heart, though: it requires knowledge
+of C++, a willingness to learn at least the basics of the `node-fibers` codebase
+and V8's isolate APIs, and skills in both API design and human interaction to
+negotiate a stable API that will meet the needs of `node-fibers` *and* that the
+V8 team feels comfortable committing to maintain. But if you're interested,
+please don't hesitate to [reach out]!
+
+[reach out]: mailto:nweiz@google.com
```
jina, line 160: are you comfortable with your email here? I've experienced spambots picking up emails in the past so I'm wondering if you want to obfuscate it (some use JS, some just do `name at emailserver dot com`.

nex3: Google's spam filters are usually pretty good. This email is publicly available in plaintext in a number of other places, so I'm not *too* worried about exposing it here.

jina: okie dokie! :)

source/blog/029-node-fibers-discontinued.md
```diff
@@ -0,0 +1,160 @@
+---
+title: "The Discontinuation of node-fibers"
+author: Natalie Weizenbaum
+date: 2021-3-26 15:00 PST
+---
+
+We have recently received the unfortunate but not entirely surprising news that
+[the `node-fibers` package has reached its end-of-life] and will not be updated
```
jathak, line 8: GitHub doesn't pick this up as a link, but I'm assuming the site itself will?

source/blog/029-node-fibers-discontinued.md
```diff
@@ -0,0 +1,160 @@
...
+
+While it's not ready for prime-time yet, the Sass team is working on a project
+called "embedded Dart Sass". This involves running Dart Sass as a *subprocess*,
+rather than a library, and communicating with it using a special protocol. This
+provides several important improvements over the existing alternatives:
+
+* Unlike running `sass` from the command line, this will still work with plugins
+  like the webpack importer. In fact, we plan to match the existing JavaScript
+  API as closely as possible. This will probably run asynchronous plugins *even
+  faster* than synchronous ones.
```
jathak, line 111: Not necessary to mention in the blog post, but I'm curious why async plugins would be faster than sync ones here (unless this is supposed to mean that async plugins on embedded Dart Sass are faster than synchronous plugins on compiled-to-JS Dart Sass?)

source/blog/029-node-fibers-discontinued.md
```diff
@@ -0,0 +1,160 @@
...
+
+* Unlike the existing JS-compiled version, this will use the Dart VM. Due to the
+  more static nature of the Dart language, the Dart VM runs Sass substantially
+  faster than Node.js, which will provide about a 2x speed improvement for large
+  stylesheets.
+
+The Node.js host for Embedded Sass is still in active development, but there's
+[a beta release] avialable (with minimal features) if you want to kick the
```
jathak, line 119:
```suggestion
[a beta release] available (with minimal features) if you want to kick the
```

jathak (approved):

source/blog/029-node-fibers-discontinued.md
nex3: They're faster because it's actually possible to run in parallel with other processing the way evented programming is *supposed* to work. In JS Dart Sass, any parallelism gains are overshadowed by the overhead from asynchrony, but Embedded Dart Sass actually does all the hard work of compilation out-of-process, so Node.js can continue to do productive work in the meantime. If the entire app is single-threaded it'll still be slower, but in a webpack-style environment where multiple builds are running at once it should be a win.

PR #553, opened 2021-06-06
Fix code examples in map and meta modules.

flachware:
Hi, I believe there is a typo in the first sass:map code example and `content` is not a valid HTML element in the first-class function example of sass:meta.

### DIFF
--- a/source/code-snippets/_example-first-class-function.html.erb
+++ b/source/code-snippets/_example-first-class-function.html.erb
@@ -46,7 +46,7 @@
 
   $fonts: Tahoma, Geneva, "Helvetica Neue", Helvetica, Arial, sans-serif
 
-  body
+  .content
     @function contains-helvetica($string)
       @return string.index($string, "Helvetica")
 
--- a/source/code-snippets/_example-first-class-function.html.erb
+++ b/source/code-snippets/_example-first-class-function.html.erb
@@ -52,7 +52,7 @@
 
     font-family: remove-where($fonts, meta.get-function("contains-helvetica"))
   ===
-  body {
+  .content {
     font-family: Tahoma, Geneva, Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


source/code-snippets/_example-first-class-function.html.erb
```diff
@@ -46,13 +46,13 @@
 
   $fonts: Tahoma, Geneva, "Helvetica Neue", Helvetica, Arial, sans-serif
 
-  content
```
Awjin, line 49: This is a valid CSS selector, but it might be clearer to explicitly make it a class selector.

source/code-snippets/_example-first-class-function.html.erb
```diff
@@ -46,13 +46,13 @@
...
+  body
```
Awjin, line 49:
```suggestion
  .content
```

source/code-snippets/_example-first-class-function.html.erb
```diff
@@ -46,13 +46,13 @@
...
     @function contains-helvetica($string)
       @return string.index($string, "Helvetica")
 
     font-family: remove-where($fonts, meta.get-function("contains-helvetica"))
   ===
-  content {
+  body {
```
Awjin, line 55:
```suggestion
  .content {
```

Awjin (changes requested):

source/code-snippets/_example-first-class-function.html.erb
flachware: It might be a valid CSS selector, but the element would not be valid HTML markup. Anyway, something more familiar like the suggested class selector makes sense to me.

Awjin (approved):
Thanks for catching these typos!

PR #556, opened 2021-06-12
Update Mixins section description

torshin:
According to the new example change variables' names in description from `transform` to `theme`, from `$property` to `$theme`.


Awjin (approved):
Good catch, thanks!

PR #559, opened 2021-06-21
Fix example on the "Slash as Division" page: divide() -> math.div()

alexander-doroshko:
According to my experiments, `divide()` doesn't work, only `math:div()` does.


Awjin:
Good catch, thanks!

PR #563, opened 2021-07-19
adding optional param description to function-exists

area73:
There is a missing parameter on  function `function-exists` parameters
https://sass-lang.com/documentation/modules/meta#function-exists.

Right now it says:
```scss
meta.function-exists($name)
function-exists($name) //=> boolean 
```

And it is suppose to be :
```scss
meta.function-exists($name, $module:null)
function-exists($name) //=> boolean 
```
In order to check for things like:
```scss
@use sass:math;

@debug meta.function-exists("div", "math")
```

### DIFF
--- a/source/documentation/modules/meta.html.md.erb
+++ b/source/documentation/modules/meta.html.md.erb
@@ -182,13 +182,21 @@ title: sass:meta
 <% end %>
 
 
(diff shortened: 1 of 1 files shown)


Awjin (approved):

Awjin:
Thanks, good catch!

PR #570, opened 2021-08-14
Remove incorrectly added SASS code from CSS snippet

samigjonov:
Link: https://sass-lang.com/documentation/syntax/special-functions#calc-clamp-element-progid-and-expression
In the compiled CSS code we can see SASS's import feature but since CSS does not support it, it needs to be removed.

Screenshot: 
image


Awjin (approved):
Thanks!

PR #586, opened 2021-10-07
Remove Sass News

jina:
I ended up shutting the project down because I was unable to maintain it.


nex3 (approved):
Pour one out... it was great while it lasted!

PR #589, opened 2021-10-11
Make the TypeDoc theme resilient against nested child lists

nex3:
It's not clear why TypeDoc generates lists of lists for JSX node
children, but it does and we need to handle it.


tool/typedoc-theme.js
```diff
@@ -68,6 +68,26 @@ class SassSiteRenderContext extends DefaultThemeRenderContext {
     return navigation;
   }, this);
 
+  // Returns the first child of a JSX node. For some reason, JSX nodes created
```
jathak, line 71: Should "first child" be "n-th child"?

PR #596, opened 2021-11-24
Add documentation explicitly describing compatibility callouts

### LINKED ISSUE #594: Please make docs for older versions easy to find
It's pretty frustrating, as a GitHub Pages user (there are *lots* of us!), to try to learn how to use SAAS from the documentation.  There's no obvious clue that I might be stuck on an older version, and once I figured that out, I was still unable to find an older version of the documentation.  FIguring out that `math.is-unitless(x)` must be spelled `unitless(x)` (especially after all the talk I found of using name prefixes—nope, `math-is-unitless` doesn't work), was really tough.  And I never know what other recent changes to the language might trip me up.

I realize that in some sense this is a problem caused for you by jekyll/GitHub, but I think saying something very explicit about it on your site and pointing to the old docs might actually increase pressure on those entities to find a way to stay current.

Thanks!

nex3:
Closes #594


nex3:
@dabarahams Is this the sort of documentation that would have been helpful for you?

source/documentation/index.html.md.erb
```diff
@@ -0,0 +1,59 @@
...
+Implementations can also be marked as "partial":
+
+<% impl_status dart: true, libsass: :partial, ruby: false do %>
+  Additional details go here.
+<% end %>
+
+This indicates that the implementation only supports some aspects of the
+feature. These compatibility indicators (and many others) have a "▶" button,
+which can be clicked to show more details about exactly how the implementations
+differ and which versions support which aspects of the feature in question.
```
Awjin, line 59: IMO this is a common enough UX pattern that we don't need to describe it.
```suggestion
This indicates that the implementation only supports some aspects of the
feature.
```

Awjin (approved):

stof:
@dabrahams is this doc helping you ? with the right mention this time.

source/documentation/index.html.md.erb
nex3: In practice, people have missed this in the past, so I think it's worth mentioning explicitly.

dabrahams:
Yes, thanks, this is wonderful!

PR #599, opened 2021-11-27
[Remove]: The extra hash that is throwing errors

Atul-Kumar-Official:


nex3:
This hash is necessary, since the syntax for the `element()` function requires it: https://developer.mozilla.org/en-US/docs/Web/CSS/element()

PR #606, opened 2022-01-14
Fix missing hyphen in private members example

querkmachine:
Tiny fix. The variable referenced inside of the mixin of the Sass private members code example does not match the variable assigned above it or referenced below it.


Awjin:
Thanks for the fix!

PR #607, opened 2022-02-01
Write a blog post announcing Embedded Sass

nex3:


source/blog/032-embedded-sass-is-live.md
```diff
@@ -0,0 +1,74 @@
...
+files especially in asynchronous mode. We expect `sass-embedded` to be a major
+boon to developers for whom compilation speed is a concern.
+
+The `sass-embedded` package fully supports the [new JS API] as well as the
+[legacy API] other than a few cosmetic options. You can use it as a drop-in
+replacement for the `sass` package, and it should work with all the same build
+plugins and libraries. Note the `sass` package is much faster at compiling files
+synchronously, the `sass-embedded` package is actually a bit faster at compiling
+them *asynchronously* instead—although for substantial Sass files it will
+generally be much faster than `sass` either way.
```
Awjin, line 26: Since there are two axes of comparison, this is a bit confusing.

```suggestion
plugins and libraries. Note that `sass-embedded` is a bit faster in
*asynchronous* mode than it is in synchronous mode (whereas the `sass`
package was faster in synchronous mode). For substantial Sass files, running
`sass-embedded` in either mode will generally be much faster than `sass`.
```

source/blog/032-embedded-sass-is-live.md
```diff
@@ -0,0 +1,74 @@
...
+
+[`sass-embedded`]: https://www.npmjs.com/package/sass-embedded
+
+Embedded Sass is an ongoing effort to make a highly-performant Sass library
+available to as many different languages as possible, starting with Node.js.
+Although Node.js already has access to the pure-JS `sass` package, the nature of
+JavaScript inherently limits how quickly this package can process large Sass
+files especially in asynchronous mode. We expect `sass-embedded` to be a major
+boon to developers for whom compilation speed is a concern.
+
```
Awjin, line 19: For the developers who are still on `node-sass`, it might be worth mentioning `node-sass` explicitly to motivate this post even more. We could briefly mention embedded is intended as an ~equally performant yet up-to-spec implementation, and endorse migration more strongly. (I guess that depends on how forward- or backward-looking you want this post to be.)

Awjin (approved):

PR #625, opened 2022-05-03
Add href to compatibility callout toggle

### LINKED ISSUE #598: Details for compatility callouts are not keyboard accessible
the toggle uses `<a class="">▶</a>` as markup. This does not make it a focusable element.

This should probably use a `<button type=button>` markup instead.

jathak:
Fixes #598.

### DIFF
--- a/helpers/sass_helpers.rb
+++ b/helpers/sass_helpers.rb
@@ -402,6 +402,7 @@ module SassHelpers
     contents << _impl_status_row('Ruby Sass', ruby) unless ruby.nil?
 
     if block_given?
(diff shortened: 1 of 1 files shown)


jathak:
@nex3: Can you deploy this branch so we can confirm it works? I'm having trouble getting the site to build locally.

nex3:
It looks like this substantially changes the look of the button, and also breaks the on-click behavior. You may need to load the site locally to debug this—let me know if I can help you debug any issues that come up.

image

nex3 (changes requested):
Can you add a comment explaining what the href is doing?

nex3 (approved):

PR #639, opened 2022-06-08
#638 fix link to Ghostlab

### LINKED ISSUE #638: Broken url under install
Hi

I noticed that in the install page there is a link to "Ghostlab".
The url should be "https://www.vanamco.com/" instead of  "https://www.vanamco.com/ghostlab/"
It is also providing a final public domain version of it now, so it should probably be marked as free.

kauefr:
Fixes #638

Fixes broken url and changes Paid to Free in the description.

Can someone still vouch for this Ghostlab software? Its current page looks kinda sketchy.

### DIFF
--- a/source/install.html.haml
+++ b/source/install.html.haml
@@ -22,8 +22,8 @@ no_container: true
         Mac
 
       %li
-        = link_to 'Ghostlab', 'http://www.vanamco.com/ghostlab/'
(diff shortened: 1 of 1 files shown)


nex3:
Given that this app's website seems to have been down since 2021, I think we should probably just get rid of this link.

PR #643, opened 2022-06-13
Update _example-mixin-arbitrary-keyword-arguments.html.erb

ivanji:
Correcting possible error in sample code


nex3:
Why do you believe this to be an error?

Issue #659, opened 2022-07-22
Document random() with units deprecation


Goodwine:
See https://github.com/sass/sass/issues/1890

- [ ] Deprecation
- [ ] Final

Goodwine:
Currently blocked on pending feedback

Goodwine:
deprecation added, now waiting on dart 2.0.0

PR #664, opened 2022-08-16
Document random($limit) behavior whe $limit has units

Goodwine:
See https://github.com/sass/sass-site/issues/659

### DIFF
--- a/data/documentation.yml
+++ b/data/documentation.yml
@@ -60,13 +60,13 @@ toc:
   - <code>sass:string</code>: /documentation/modules/string
 - Breaking Changes: /documentation/breaking-changes
   :children:
+  - Random With Units: /documentation/breaking-changes/random-with-units
   - Invalid Combinators: /documentation/breaking-changes/bogus-combinators
   - Media Queries Level 4: /documentation/breaking-changes/media-logic
   - <code>/</code> as Division: /documentation/breaking-changes/slash-div
   - Color Units: /documentation/breaking-changes/color-units
   - Extending Compound Selectors: /documentation/breaking-changes/extend-compound
   - CSS Variable Syntax: /documentation/breaking-changes/css-vars
-  - Random With Units: /documentation/breaking-changes/random-with-units
 - Command Line: /documentation/cli
   :children:
   - Dart Sass: /documentation/cli/dart-sass
--- a/data/documentation.yml
+++ b/data/documentation.yml
@@ -66,6 +66,7 @@ toc:
   - Color Units: /documentation/breaking-changes/color-units
   - Extending Compound Selectors: /documentation/breaking-changes/extend-compound
   - CSS Variable Syntax: /documentation/breaking-changes/css-vars
+  - Random With Units: /documentation/breaking-changes/random-with-units
 - Command Line: /documentation/cli
   :children:
   - Dart Sass: /documentation/cli/dart-sass
--- a/source/documentation/breaking-changes.html.md.erb
+++ b/source/documentation/breaking-changes.html.md.erb
@@ -37,3 +37,6 @@ These breaking changes are coming soon or have recently been released:
 
 * [The syntax for CSS custom property values changed](breaking-changes/css-vars)
   in Dart Sass 1.0.0, LibSass 3.5.0, and Ruby Sass 3.5.0.
+
+* [`random($limit)` with units](breaking-changes/random-with-units) beginning in
+  Dart Sass 1.54.5.
--- a/source/documentation/breaking-changes.html.md.erb
+++ b/source/documentation/breaking-changes.html.md.erb
@@ -23,6 +23,9 @@ time-sensitive, so they may be released with new minor version numbers instead.
 
 These breaking changes are coming soon or have recently been released:
 
+* [`random()` with units](breaking-changes/random-with-units) beginning in Dart
+  Sass 1.54.5.
+
 * [Selectors with invalid combinators will be
   invalid](breaking-changes/bogus-combinators) beginning in Dart Sass 1.54.0.
 
@@ -37,6 +40,3 @@ These breaking changes are coming soon or have recently been released:
 
 * [The syntax for CSS custom property values changed](breaking-changes/css-vars)
   in Dart Sass 1.0.0, LibSass 3.5.0, and Ruby Sass 3.5.0.
-
-* [`random($limit)` with units](breaking-changes/random-with-units) beginning in
-  Dart Sass 1.54.5.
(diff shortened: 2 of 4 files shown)


data/documentation.yml
```diff
@@ -1,74 +1,75 @@
 toc:
```
nex3, line 1: Why is this file changing so much?

source/documentation/breaking-changes.html.md.erb
```diff
@@ -37,3 +37,6 @@ These breaking changes are coming soon or have recently been released:
 
 * [The syntax for CSS custom property values changed](breaking-changes/css-vars)
   in Dart Sass 1.0.0, LibSass 3.5.0, and Ruby Sass 3.5.0.
+
+* [`random($limit)` with units](breaking-changes/random-with-units) beginning in
```
nex3, line 41: To match the order of other changes here, more recent changes should go at the top.

source/documentation/modules/math.html.md.erb
```diff
@@ -540,6 +540,20 @@ title: sass:math
   If `$limit` is a number greater than or equal to 1, returns a random whole
   number between 1 and `$limit`.
 
+  <% heads_up do %>
+    `random($limit)` ignores units in `$limit`. [This behavior will be
+    deprecated] and `random($limit)` will return a random integer with the same
```
nex3, line 545:
```suggestion
    `random($limit)` ignores units in `$limit`. [This behavior is
    deprecated] and in the future `random($limit)` will return a random integer with the same
```

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
+---
+title: "Breaking Change: Random With Units"
+introduction: >
+  Sass `random($limit)` will no longer ignore `$limit` units in a future
+  version. Passing a `$limit` with units will be deprecated to ensure the
```
nex3, line 5:
```suggestion
  version. Passing a `$limit` with units is deprecated to ensure the
```

This will be landing approximately concurrently with the deprecation warning, so we should use present tense. Same goes elsewhere in the documentation.

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
+---
+title: "Breaking Change: Random With Units"
+introduction: >
+  Sass `random($limit)` will no longer ignore `$limit` units in a future
```
nex3, line 4:
```suggestion
  The `random()` function will no longer ignore units on the `$limit` argument in a future
```

* We try to use "the NAME TYPE" when referring to Sass constructs, including functions and parameters of those functions. This has three benefits:

  * It allows us to avoid starting sentences with an uncapitalized code name.
  * If we're linking to those constructs' documentation, it means the link covers more than just the `<code>` tag which makes it more visible to readers.
  * It gives unfamiliar readers more context to hook onto.

* We usually refer to functions in documentation without parameters, since parameter lists get quite long for more complex functions like `color.scale()`.

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
...
+  version. Passing a `$limit` with units will be deprecated to ensure the
+  compilation results don't change once the new behavior lands.
+---
+
+The function [`random($limit)`] has historically ignored units in `$limit` which
```
nex3, line 9: "The `random()` function", as above.

nex3, line 9:
```suggestion
The function [`random($limit)`] has historically ignored units in `$limit` and
```

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
...
+returned a unitless value. For example `random(100px)` would drop "px" and
+return a value like `42`.
+
+A future implementation of Dart Sass will stop ignoring units for the `$limit`
+argument will return a random integer with the same units.
```
nex3, line 14:
```suggestion
argument and return a random integer with the same units.
```

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
...
+  @debug math.random(100px)  // 42px
+<% end %>
+
+## Transition Period
+
+<% impl_status dart: '1.54.5', libsass: false, ruby: false %>
+
+First, we'll emit deprecation warnings for the previous uses of `random()` where
+the `$limit` argument has units. We'll suggest a way to preserve the existing
+behavior and a way that emulates the future implementation.
```
nex3, line 32: "We" is confusing here. Is "we" the documentation author? The Sass team? The Sass compiler?

I'd say "The Sass compiler" explicitly.

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,36 @@
...
+
+In Dart 2.0.0 passing a `$limit` number with units will be an error, and then in
+a future minor release (i.e. >=2.1.0) we'll update `random($limit)` so the
```
nex3, line 35: I think listing an explicit version here is likely to make people think that's the version we'll start supporting units in, even if you say otherwise. I'd just leave it off.

Also, for future reference, I've heard the advice to avoid abbreviations like "i.e." because they're unfamiliar to readers who don't come from a Romance language background.

nex3 (changes requested):

data/documentation.yml
Goodwine: oh I didn't realize, it was an autoformatter lol

source/documentation/breaking-changes/random-with-units.md.erb
Goodwine: thanks for the feedback, I tend to over-use `i.e.` and `e.g.`, I'll keep this in mind :)

source/documentation/breaking-changes/random-with-units.md.erb
```diff
@@ -0,0 +1,37 @@
+---
+title: "Breaking Change: Random With Units"
+introduction: >
+  The `random()` function will no longer ignore units on the `$limit` argument
+  in a future version. Passing a `$limit` with units is deprecated to ensure the
+  compilation results don't change once the new behavior lands.
+---
+
+The function [`random()`] has historically ignored units in `$limit` and
```
nex3, line 9:
```suggestion
[The `random()` function] has historically ignored units in `$limit` and
```

nex3 (approved):
Once change, then looks good!

Issue #676, opened 2022-09-28
Incorrect comment on document's 'Equality' part


custardcream98:
https://sass-lang.com/documentation/operators/equality

In the codeblock below, it says `@debug 1 != 1px; // true` but I've found that it's not.

We should correct this. 😁

PR #677, opened 2022-09-28
fix: Add More Info on Equality Operators page

custardcream98:
# Issue

Resolves: #676

# Summery

`@debug 1 != 1px;` is `false`, and `@debug 1 != 1rem;` is also `false`.

So I've added some and fixed comment.

### DIFF
--- a/source/documentation/operators/equality.html.md.erb
+++ b/source/documentation/operators/equality.html.md.erb
@@ -18,10 +18,11 @@ they're the same type *and* the same value, which means different things for
 different types:
 
 [expressions]: ../syntax/structure#expressions
+[deprecated]: /blog/libsass-is-deprecated
 
 * [Numbers][] are equal if they have the same value *and* the same units, or if
   their values are equal when their units are converted between one another.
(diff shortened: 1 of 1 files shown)


custardcream98:
I've found out that it wasn't a typo, so I've added some more for those who get confused like me. 😂😂

custardcream98:
So,

```scss
@debug 1 != 1px;
```

is `false` on LibSass version under 3.6.5
and `true` on LibSass version 3.6.5 thanks to [this change](https://github.com/sass/libsass/pull/3095)
(and also `true` on lateset Dart Sass)

nex3:
Thanks for the contribution, but we've already got info about the change in equality behavior on that page, in our standardized component for describing cross-version/-implementation differences:

https://github.com/sass/sass-site/blob/02771170cf46382d33d0e5ca558135ea0d81ffbe/source/documentation/operators/equality.html.md.erb#L5-L11

custardcream98:
Oh, I didn't notice that toggle button. Thanks! 👍

PR #678, opened 2022-09-28
Docs:  Edit [numeric.html.md.erb]

khv2644511:
Document is not correct, 'Numeric Operators' part

(https://sass-lang.com/documentation/operators/numeric) In the code block below, 
it says `@debug 5px * 3px; // 15px*px` but It's actually printed 'error' message.

Resolves: #675


jathak:
See my comment on #675.

PR #681, opened 2022-10-17
Clarify flow control scope documentation

### LINKED ISSUE #669: Sass: Variables (Documentation)
I am not sure if I should write this issue here, but here we go! 
image
as you see in this picture from the documentation of sass [here](https://sass-lang.com/documentation/variables) about the variables. It says that flow control statements can't create/declare new variables, which is not true

```
@if 1 {
 $index: 20px;
 @debug $index;
}
```

```
@for $i from 1 through 4 {
  $index: 20px;
  @debug $index;
}
```
In both cases above, the variables are declared and debugged without prompting any errors, as long as I am using the variables inside the flow control.. maybe you should specify that in the documentation because it creates some confusion for beginners like me. have a great day!

jathak:
Fixes #669.

### DIFF
--- a/source/documentation/variables.html.md.erb
+++ b/source/documentation/variables.html.md.erb
@@ -279,9 +279,9 @@ assign a value to a variable, or build up a value as part of a loop.
 
 <% heads_up do %>
   Variables in flow control scope can assign to existing variables in the outer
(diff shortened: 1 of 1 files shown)


stof:
@nex3 what is the status of this documentation update ?

nex3 (approved):

PR #682, opened 2022-10-19
Fixing typo

dylakim:
Noticed a typo in the At-Rules/@at-root docs. I believe the second "includes" is meant to be "excludes".


nex3:
Thanks!

PR #689, opened 2022-12-09
Blog: announce the tar file permissions vulnerability

nex3:
See google/dart_cli_pkg#109


jathak (approved):

nex3:
I've asked an internal Google security team to spot-check this, so I'm waiting to land it until they do.

ntkme:
@nex3 Can we also mention community maintained sass-embedded ruby gem as being impacted? Specifically, all past versions (<=1.56.2) but only “non-native” ruby gems installed as root are affected.

Issue #693, opened 2023-02-20
Error in the syntax documentation


Debjyoti0404:
In case of sass, we do not need to put semicolon ( ; ) before terminating to new line. But in the documentation, it is not mentioned properly as we can see it was written **it uses semicolon to describe the format of the document** . We can see that in the attached screenshot.
![Screenshot_20230221_023308](https://user-images.githubusercontent.com/103656712/220199557-1de50c5e-fb85-4a40-9ecc-a4f7e0e1b1e3.png)

The issue was in the syntax.html.md.erb file. **File path : sass-site/source/documentation/syntax.html.md.erb**
![Screenshot_20230221_023434](https://user-images.githubusercontent.com/103656712/220199358-9ccf4670-8c60-45ed-be50-16a19ca7df14.png)

jathak:
I think this is just a misinterpretation of the sentence in question; it's saying that the indented syntax uses indentation instead of _both_ curly braces and semicolons (that is, the SCSS syntax uses "curly braces and semicolons" instead of indentation".

PR #694, opened 2023-02-20
#693 fixed sass syntax documentation

Debjyoti0404:
Error in the documentation regarding usage of semicolon in sass syntax documentation is fixed.


jathak:
See my response to #693.

PR #702, opened 2023-03-09
Document --fatal-deprecation

jathak:
For future reference, here is the Markdown-equivalent of the table below, so that we can replace the HTML table with it once we migrate to Eleventy

```
| ID                    | Version | Description                                                    |
|-----------------------|---------|----------------------------------------------------------------|
| `call-string`         | 0.0.0   | Passing a string directly to `meta.call()`                     |
| `elseif`              | 1.3.2   | Using `@elseif` instead of `@else if`                          |
| [`moz-document`]      | 1.7.2   | Using `@-moz-document` (except for the empty url prefix hack). |
| `new-global`          | 1.17.2  | Declaring new variables with `!global`.                        |
| `color-module-compat` | 1.23.0  | Using color module functions in place of plain CSS functions.  |
| [`slash-div`]         | 1.33.0  | Using the `/` operator for division.                           |
| [`bogus-combinators`] | 1.54.0  | Leading, trailing, and repeated combinators.                   |
| [`strict-unary`]      | 1.55.0  | Ambiguous `+` and `-` operators.                               |
| [`function-units`]    | 1.56.0  | Passing invalid units to built-in functions.                   |

[`moz-document`]: breaking-changes/moz-document
[`slash-div`]: breaking-changes/slash-div
[`bogus-combinators`]: breaking-changes/bogus-combinators
[`strict-unary`]: breaking-changes/strict-unary
[`function-units`]: breaking-changes/function-units
```

### DIFF
--- a/source/documentation/breaking-changes.html.md.erb
+++ b/source/documentation/breaking-changes.html.md.erb
@@ -40,3 +40,75 @@ These breaking changes are coming soon or have recently been released:
 
 * [The syntax for CSS custom property values changed](breaking-changes/css-vars)
   in Dart Sass 1.0.0, LibSass 3.5.0, and Ruby Sass 3.5.0.
+
+## Early Opt-In
+
+<% impl_status dart: '1.59.0', libsass: false, ruby: false %>
+
+Users who wish to treat certain deprecation warnings as errors immediately can
+do so with the `--fatal-deprecation` command line option. For example,
+`--fatal-deprecation=slash-div` will treat all deprecation warnings for
+`/`-as-division as errors.
+
+<table style="width:100%">
+<thead>
+  <tr style="text-align: left">
+    <th>ID</th>
+    <th>Version</th>
+    <th>Description</th>
+  </tr>
+</thead>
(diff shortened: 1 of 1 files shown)


source/documentation/breaking-changes.html.md.erb
```diff
@@ -40,3 +40,75 @@ These breaking changes are coming soon or have recently been released:
 
 * [The syntax for CSS custom property values changed](breaking-changes/css-vars)
   in Dart Sass 1.0.0, LibSass 3.5.0, and Ruby Sass 3.5.0.
+
+## Early Opt-In
```
nex3, line 44: It's probably better to document this on https://sass-lang.com/documentation/cli/dart-sass and link to it from here.

nex3 (changes requested):

source/documentation/breaking-changes.html.md.erb
jathak: Done

nex3 (approved):

PR #706, opened 2023-03-22
Fix redirects for random-with-units and color-units

### LINKED ISSUE #704: Misbehaving link
The following link:

> https://sass-lang.com/documentation/breaking-changes/random-with-units

...which is found under the **breaking changes** section here:

image

...doesn't work as expected. The HTTP response comes back with a `content-type` header value of `application/octet-stream` which means **the browser opens a File Download dialog rather than loading the page**:

image

Goodwine:
Fixes #704

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -72,8 +72,10 @@ Dir['source/documentation/breaking-changes/**'].each do |file|
 end
 
 for url in %w[d/random-with-units documentation/breaking-changes/random-with-units
-    d/breaking-changes/color-units documentation/breaking-changes/color-units] do
-  redirect url, to: "/documentation/breaking-changes/function-units"
+    d/color-units documentation/breaking-changes/color-units] do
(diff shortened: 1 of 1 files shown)


config.rb
```diff
@@ -72,8 +72,10 @@
 end
 
 for url in %w[d/random-with-units documentation/breaking-changes/random-with-units
-    d/breaking-changes/color-units documentation/breaking-changes/color-units] do
-  redirect url, to: "/documentation/breaking-changes/function-units"
+    d/color-units documentation/breaking-changes/color-units] do
+  # Middleman requires the origin URL to contain an extension, otherwise it will
```
nex3, line 76: I think it's GitHub pages' behavior that's causing this, since we don't run Middleman as a server in production.

nex3 (approved):

PR #711, opened 2023-03-29
Add a blog post about native CSS nesting

nex3:
See https://github.com/sass/sass/issues/3030

### DIFF
--- a/source/blog/036-sass-and-native-nesting.md
+++ b/source/blog/036-sass-and-native-nesting.md
@@ -129,14 +129,15 @@ that's fully compatible with existing Sass behavior.
 We _will_ add support for parsing plain CSS nesting in `.css` files. This
 nesting won't be resolved in any way; Sass will just emit it as-is.
 
-**In the long term**, once `:is()` is supported by 98% of the global browser
-market share, we'll start transitioning Sass to emit `:is()` when resolving Sass
-nesting. This will make Sass behave like CSS in the first two behavioral
+**In the long term**, once [`:is()` is supported by 98% of the global browser
+market share], we'll start transitioning Sass to emit `:is()` when resolving
+Sass nesting. This will make Sass behave like CSS in the first two behavioral
 incompatibilities. We will consider this a breaking change, and release it as
 part of a major version release to avoid unexpectedly breaking existing
 stylesheets. We'll do our best to make this transition as smooth as possible
 using the [Sass Migrator].
 
+[`:is()` is supported by 98% of the global browser market share]: https://caniuse.com/css-matches-pseudo
 [Sass Migrator]: /documentation/cli/migrator
 
 We will _not_ drop our current behavior for `&-suffix` unless we can come up
--- a/source/blog/036-sass-and-native-nesting.md
+++ b/source/blog/036-sass-and-native-nesting.md
@@ -79,7 +79,7 @@ nesting**. This affects three different cases:
    but Sass will not, since the element matching `.foo` is outside the element
    matching `.green-theme`.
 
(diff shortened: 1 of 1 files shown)


source/blog/036-sass-and-native-nesting.md
```diff
@@ -0,0 +1,145 @@
...
+       <p class=bar>
+     </div>
+   </div>
+   ```
+
+   but Sass will not, since the element matching `.foo` is outside the element
+   matching `.green-theme`.
+
+3. Native CSS nesting and Sass nesting both support syntax that looks like
+   `&foo`, but it means different things. In native CSS, this adds a type
```
Goodwine, line 83: Nitpicking - the 2 bullets above talk about "Sass does X, native CSS does Y" but this bullet changes the order to explain what native CSS does first and then Sass, I found that to be slightly confusing

source/blog/036-sass-and-native-nesting.md
```diff
@@ -0,0 +1,145 @@
...
+   selector to the parent selector, so
+   
+   ```scss
+   .foo {
+     &div { /* ... */ }
+   }
+   ```
+
+   produces the selector `div.foo`. But in Sass, this adds a suffix to the
```
Goodwine, line 92: I think you should add what Sass would do here: `.foodiv` to highlight how different it behaves. The example below does show that, but I think it's not clear enough because the inputs are different

Goodwine (changes requested):

source/blog/036-sass-and-native-nesting.md
```diff
@@ -0,0 +1,145 @@
...
+First of all, we won't ever change existing valid Sass code so that it starts
+emitting CSS that's incompatible with widely-used browsers. This means that even
+if we did decide to phase out Sass nesting and just emit plain CSS nesting
+instead, we wouldn't do so until [98% of the global browser market share]
+supported native nesting.
+
+[98% of the global browser market share]: https://github.com/sass/dart-sass#browser-compatibility
+
+More importantly, though, **native CSS nesting is subtly incompatible with Sass
+nesting**. This effects three different cases:
```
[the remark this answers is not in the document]
nex3: no, you're right :)

source/blog/036-sass-and-native-nesting.md
```diff
@@ -0,0 +1,145 @@
...
+   descendant combinators will behave differently.
+
+   ```scss
+   .foo .bar {
+     .green-theme & { /* ... */ }
+   }
+   ```
+
+   produces the selector `.green-theme .foo .bar` in Sass, but in native CSS it
+   produces `.green-theme :is(.foo .bar)`. This means that the native CSS
```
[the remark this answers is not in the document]
nex3: Yeah it's bad. I think authors are going to be very unhappy about this behavior, but the CSSWG says matching Sass's behavior is "too weird with sibling combinators".

jathak (approved):

source/blog/036-sass-and-native-nesting.md
```diff
@@ -79,27 +79,27 @@ nesting**. This effects three different cases:
    but Sass will not, since the element matching `.foo` is outside the element
    matching `.green-theme`.
 
-3. Native CSS nesting and Sass nesting both support syntax that looks like
-   `&foo`, but it means different things. In native CSS, this adds a type
-   selector to the parent selector, so
-   
+3. Sass nesting and native CSS nesting and both support syntax that looks like
```
Goodwine, line 82: Extra `and` after `nesting`

Goodwine (approved):

Issue #660, opened 2022-07-30
maven cli wrapper plugin


ennerf:
Reading through the documentation I wasn't sure how to best integrate with Maven, so I wrote a small plugin that downloads and calls the CLI: 

[sass-cli-maven-plugin](https://github.com/HebiRobotics/sass-cli-maven-plugin)

Feel free to add it to the docs, or ignore if there is already a better way to do it.

Goodwine:
There was a Maven plugin for libsass but not for dart-sass, thanks for letting us know, I'll add a link shortly :)

melloware:
There is a full proper wrapper now using Dart Sass with Maven or Gradle.

https://github.com/cleydyr/dart-sass-maven-plugin

melloware:
I am pretty sure the one I posted supports all 3 OS Windows Linux MacOS and knows also has a Sass Watch properly integrated.  I didn't study the first one in detail to compare the two but the latter emulates was the old lib sass plug-in did almost exactly.

ennerf:
> the one I posted supports all 3 OS Windows Linux MacOS and knows which you are using. also has a Sass Watch properly integrated.

The first one does as well. I had a quick look at the source code and I think the main differences are:

**sass-cli-maven-plugin**
* supports arbitrary versions of the dart plugin and downloads the specified version for the host OS
* cli arguments are specified as an`<args>` list and passed to the executable, e.g., `<arg>--embedd-sources</arg>`
* currently the only argument with special support is `watch`, but it'd be fairly straight forward to add maven parameter wrappers
* the reasoning behind this approach was to support new versions/features immediately without requiring a plugin release

**dart-sass-maven-plugin**
* bundles a fixed version of dart-sass in the jar and extracts it at runtime
* the arguments are exposed as maven parameters, i.e., `<embedSources>true</<embedSources>`
* the bundled approach requires matching plugin releases for new sass-dart releases

Basically one is a thin wrapper around the CLI executable, and the other one is an abstraction on top that hides the actual calls.  It looks like the dart-sass plugin was started a bit earlier, so I wouldn't have written another one if I had been aware of it. That being said, I don't like that the latest plugin release is already at least 9 months worth of dart-sass releases out of date.

melloware:
[...]

PR #714, opened 2023-04-03
Add link to a new Maven plugin on the block.

Goodwine:
Document the differences between both Maven plugins

image

Fixes https://github.com/sass/sass-site/issues/660

### DIFF
--- a/source/install.html.haml
+++ b/source/install.html.haml
@@ -45,18 +45,31 @@ no_container: true
 
     %ul
       %li
-        #{link_to 'Java', 'https://mvnrepository.com/artifact/de.larsgrefer.sass'},
-        including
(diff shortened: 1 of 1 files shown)


source/install.html.haml
```diff
@@ -44,20 +44,33 @@ no_container: true
...
-        plugins.
-
       %li
         #{link_to 'Ruby', 'https://github.com/ntkme/sass-embedded-host-ruby#readme'}
 
       %li
         #{link_to 'Swift', 'https://github.com/johnfairh/swift-sass#readme'}
 
+      %li
+        #{link_to 'Java', 'https://mvnrepository.com/artifact/de.larsgrefer.sass'},
```
Goodwine, line 54: I moved the links to Java to the bottom because with the sublist it kinda obscured the other two options. They are no longer in alphabetical order, not sure how much that matters in this case.

nex3 (approved):

PR #729, opened 2023-05-19
Add a blog post about the new Embedded Protocol proposal

nex3:


source/blog/036-rfc-embedded-protocol-2.md
```diff
@@ -0,0 +1,76 @@
...
+We're using this opportunity to also introduce three much smaller breaking
+changes:
+
+1. The specification for the embedded protocol and the protocol buffer
+   definition have been moved to [the Sass language repository] so that they can
+   be updated at the same time as changes to the language and the JS API.
+
+   [the Sass language repository]: https://github.com/sass/sass/blob/main/spec/embedded-protocol.md
+
+2. The embedded protocol now explicitly declares fields as `optional` using the
```
Goodwine, line 35: Nit:
> The embedded protocol now explicitly declares which fields are `optional` ..

Or something like that? Otherwise when I read it it comes across as if "everything is optional now"

source/blog/036-rfc-embedded-protocol-2.md
```diff
@@ -0,0 +1,76 @@
...
+will have the full Dart Sass executable available to them at native Dart VM
+speeds.
+
+This also helps simplify the Sass team's organization by reducing the number of
+separate repositories and release processes we need to manage.
+
+## Wire-Level Compilation ID
+
+We're pulling the compilation ID out to the protocol level in order to provide
+better parallelism, particularly on the side of the embedded compiler. Sass
```
Goodwine, line 64: Nit: Should this say "concurrency" rather than "parallelism"? (if so same on line 75)

Because the change makes concurrency easier, whether it's parallel or not is up to the users, no?

I may be misremembering, but IIRC concurrency is more correct here, or did I get them the other way around? 😆

Goodwine (approved):
couple of suggestions, but lgtm

PR #738, opened 2023-06-15
Update calculation functions documentation

pamelalozano16:
Context: [Calc functions proposal](https://github.com/sass/sass/blob/main/accepted/calc-functions.md)
Fix: https://github.com/sass/dart-sass/pull/1970
Tests: https://github.com/sass/sass-spec/pull/1912

### DIFF
--- a/source/documentation/values/calculations.html.md.erb
+++ b/source/documentation/values/calculations.html.md.erb
@@ -6,7 +6,7 @@ introduction: >
   much as possible, even if they're combined with one another.
 ---
 
-<% impl_status dart: '1.40.0', libsass: false, ruby: false do %>
+<% impl_status feature: 'calc, clamp, min, and max', dart: '1.40.0', libsass: false, ruby: false do %>
   LibSass, Ruby Sass, and versions of Dart Sass prior to 1.40.0 parse `calc()`
   as a [special function] like `element()`.
 
@@ -18,6 +18,16 @@ introduction: >
   function] like `element()`.
 
   [plain CSS function]: ../at-rules/function#plain-css-functions
+<% end %>
+
+<% impl_status feature: 'Other functions', dart: '1.40.0', libsass: false, ruby: false do %>
+  LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
+  the following [math functions] as Sass functions. To create a plain CSS calculation function
+  call for those implementations, you can write something like
+  `sqrt(#{$number})` instead.
+
+  [special functions]: ../syntax/special-functions
+  [math functions]: https://www.w3.org/TR/css-values-4/#math
 <% end %>
 
 <% example(autogen_css: false) do %>
@@ -161,22 +171,55 @@ CSS to unitless numbers:
   @debug calc(-infinity) < math.$min-number  // true
 <% end %>
 
-## Calculation Functions
+## [Calculation Functions]
+ 
+* Comparison Functions: [`min()`], [`max()`], and [`clamp()`]
+* Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`], [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
+* Sign-Related Functions: [`abs()`] and [`sign()`].
 
-CSS added support for [calculation functions] in Values and Units Level
+[Calculation Functions]: https://www.w3.org/TR/css-values-4/#math
+[`min()`]: ../modules/math#min
+[`max()`]: ../modules/math#max
+[`clamp()`]: ../modules/math#clamp
+[`round()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/round
+[`sin()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sin
+[`cos()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/cos
+[`tan()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/tan
+[`asin()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/asin
+[`acos()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/acos
+[`atan()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/atan
+[`atan2()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/atan2
+[`pow()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/pow
+[`sqrt()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sqrt
+[`hypot()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/hypot
+[`log()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/log
+[`abs()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/abs
+[`exp()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/exp
+[`mod()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/mod
+[`rem()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/rem
+[`sign()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sign
+
+## [Special Cases]
+
+CSS added support for [`min()` and `max()` functions] in Values and Units Level
 4, from where they were quickly adopted by Safari [to support the iPhoneX]. But
-Sass supported its own [math functions] long before this, and
+Sass supported its own [`round()`], [`abs()`], [`min()`] and [`max()`] long before this, and
 it needed to be backwards-compatible with all those existing stylesheets. This
 led to the need for extra-special syntactic cleverness.
 
-[calculation functions]: https://drafts.csswg.org/css-values-4/#calc-notation
+[`min()` and `max()` functions]: https://www.w3.org/TR/css-values-4/#comp-func
 [to support the iPhoneX]: https://webkit.org/blog/7929/designing-websites-for-iphone-x/
-[math functions]: ../modules/math
+[`round()`]: ../modules/math#round
+[`abs()`]: ../modules/math#abs
+[`min()`]: ../modules/math#min
+[`max()`]: ../modules/math#max
 
 If a calculation function call is a valid calculation expression, it will
 be parsed as a calculation. But as soon as any part of the call contains a
 SassScript feature that isn't supported in a calculation, like the [modulo
-operator], it's parsed as a call to Sass's core calculation function
+operator], it's parsed as a call to the appropriate Sass math
 instead.
 
 Since calculations are simplified to numbers when possible anyway, the only
@@ -187,11 +230,13 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
 
 <% heads_up do %>
   Other calculations don't allow unitless numbers to be added to, subtracted
-  from, or compared to numbers with units. `round()`, `abs()`, `min()`, and `max()` are different,
-  though: for backwards-compatibility with the global Sass `round()`, `abs()`, `min()`, and `max()`
+  from, or compared to numbers with units. `min()` and `max()` are different,
+  though: for backwards-compatibility with the global Sass `min()` and `max()`
   functions which allow unit/unitless mixing for historical reasons, these units
-  can be mixed as long as they're contained directly within a `round()`, `abs()`, `min()`, or `max()`
+  can be mixed as long as they're contained directly within a `min()` or `max()` 
   calculation.
+  For instance, `sin($number: 4)` will throw an error, as `sin()` was 
+  never a global Sass function.
 <% end %>
 
 <!-- TODO(nweiz): auto-generate this CSS once we're compiling with Dart Sass -->
@@ -264,7 +309,7 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
   LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
   `round()` a as Sass function. To create a plain CSS calculation function
   call for those implementations, you can write something like
-  `unquote("round(#{$number})")` instead.
+  `round(#{$number})` instead.
 
   [special functions]: ../syntax/special-functions
 <% end %>
@@ -310,7 +355,7 @@ a value to be rounded and a rounding interval `step`.
   LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
   `abs()` a as Sass function. To create a plain CSS calculation function
   call for those implementations, you can write something like
-  `unquote("abs(#{$number})")` instead.
+  `abs(#{$number})` instead.
 
   [special functions]: ../syntax/special-functions
 <% end %>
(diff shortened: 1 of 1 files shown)


source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +161,69 @@ CSS to unitless numbers:
...
 <% end %>
 
-## `min()` and `max()`
+## Calculation Functions
 
 <% impl_status dart: ">=1.11.0 <1.42.0", libsass: false, ruby: false, feature: "Special function syntax" do %>
   LibSass, Ruby Sass, and versions of Dart Sass prior to 1.11.0 *always* parse
-  `min()` and `max()` as Sass functions. To create a plain CSS `min()` or
-  `max()` call for those implementations, you can write something like
+  `round()`, `abs()`, `min()`, and `max()` as Sass functions. To create a plain CSS calcultion function
```
pamelalozano16, line 168: Calculation*

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +161,69 @@ CSS to unitless numbers:
...
+* Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`], [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
+* Sign-Related Functions: [`abs()`] and [`sign()`].
+
 [`min()`]: ../modules/math#min
 [`max()`]: ../modules/math#max
-
-If a `min()` or `max()` function call is a valid calculation expression, it will
+[`clamp()`]: ../modules/math#clamp
```
pamelalozano16, line 204: Link all MDM

source/documentation/values/calculations.html.md.erb
```diff
@@ -207,15 +234,18 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+  though: for backwards-compatibility with the global Sass `round()`, `abs()`, `min()`, and `max()`
   functions which allow unit/unitless mixing for historical reasons, these units
-  can be mixed as long as they're contained directly within a `min()` or `max()`
+  can be mixed as long as they're contained directly within a `round()`, `abs()`, `min()`, or `max()`
   calculation.
 <% end %>
 
 <!-- TODO(nweiz): auto-generate this CSS once we're compiling with Dart Sass -->
 
+### Backwards-compatible with global Sass functions.
```
pamelalozano16, line 246: Delete this title

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +258,117 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+    padding-right: math.abs(-7.5%)
+
+  ===
+  .post {
+    padding-left: 10px;
+    padding-right: 7.5%;
+  }
+<% end %>
+
+### Other math functions
```
jathak, line 340: This section should be earlier in the article, so the exceptional cases are listed afterwards.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +258,117 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+<% impl_status dart: "1.64.0", libsass: false, ruby: false, feature: "Special function syntax" do %>
+  LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
+  the following functions as Sass functions. To create a plain CSS calculation function
+  call for those implementations, you can write something like
+  `unquote("sqrt(#{$number})")` instead.
```
jathak, line 345: This impl status should go immediately below the one for first-class calc in 1.40.0 and read similarly. Except for `min`, `max`, `round`, and `abs`, the other functions were never Sass global functions, so there's no backwards compatibility to maintain. Doing this unquote trick is unnecessary, since Sass just defaults to emitting a CSS function if there's no Sass function with that name. 1.64.0 is just the first version where these will simplify during Sass compilation.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +161,22 @@ CSS to unitless numbers:
   @debug calc(-infinity) < math.$min-number  // true
 <% end %>
 
-## `min()` and `max()`
+## Calculation Functions
```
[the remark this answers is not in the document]
pamelalozano16: Good point, although @nex3 mentioned in our meeting that we shouldn't talk about past events (I had a title mentioning `Backwards-compatible with global Sass functions.`) and I'm not sure if this applies to that.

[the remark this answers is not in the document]
jathak: Okay, in that case, maybe a better name for this section is "Special Cases"

source/documentation/values/calculations.html.md.erb
pamelalozano16: Okay I'll move it before the `Special Cases` section and within a section titled `Calculation functions` with a link to  [ https://www.w3.org/TR/css-values-4/#math](https://www.w3.org/TR/css-values-4/#math). Does that sound ok?

source/documentation/values/calculations.html.md.erb
pamelalozano16: Sounds good, just added a `impl_status` under the introduction title

source/documentation/values/calculations.html.md.erb
```diff
@@ -20,6 +20,16 @@ introduction: >
...
 
+<% impl_status feature: 'Other functions', dart: '1.40.0', libsass: false, ruby: false do %>
+  LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
+  the following [math functions] as Sass functions. To create a plain CSS calculation function
+  call for those implementations, you can write something like
+  `sqrt(#{$number})` instead.
+
+  [special functions]: ../syntax/special-functions
+  [math functions]: https://www.w3.org/TR/css-values-4/#math
+<% end %>
```
nex3, line 31: This seems to be redundant with the `impl_status`es for the `abs()` and `round()` sections below.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
   @debug calc(-infinity) < math.$min-number  // true
 <% end %>
 
-## `min()` and `max()`
+## [Calculation Functions]
```
nex3, line 174: This needs an `impl_status` that indicates that the new functions were added in 1.64.0.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
...
+ 
+* Comparison Functions: [`min()`], [`max()`], and [`clamp()`]
```
nex3, line 176: It would be good to have some brief prose here to explain what's being listed (such as "Sass parses the following functions as calculations").

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
...
+[`sqrt()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sqrt
+[`hypot()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/hypot
+[`log()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/log
+[`abs()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/abs
+[`exp()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/exp
+[`mod()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/mod
+[`rem()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/rem
+[`sign()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sign
+
+## [Special Cases]
```
nex3, line 204: "Special Cases" doesn't communicate very much—it doesn't gie the reader any hint which cases are special, or what's special about them. I'd call this something like "Legacy Global Functions".

nex3, line 204: This should probably be a third-level header 9and everything beneath it fourth-level), since it's conceptually part of "Calculation Functions".

nex3, line 204: Headers generally shouldn't contain links.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
...
-  was backwards-incompatible with the global `min()` and `max()` functions, so
-  that behavior was reverted.
-
-  [special functions]: ../syntax/special-functions
-<% end %>
+[Calculation Functions]: https://www.w3.org/TR/css-values-4/#math
+[`min()`]: ../modules/math#min
+[`max()`]: ../modules/math#max
+[`clamp()`]: ../modules/math#clamp
+[`round()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/round
```
nex3, line 186: I think these should link to their sections of this document below.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
...
+[`mod()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/mod
+[`rem()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/rem
+[`sign()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sign
+
+## [Special Cases]
 
 CSS added support for [`min()` and `max()` functions] in Values and Units Level
 4, from where they were quickly adopted by Safari [to support the iPhoneX]. But
-Sass supported its own [`min()`] and [`max()`] functions long before this, and
+Sass supported its own [`round()`], [`abs()`], [`min()`] and [`max()`] long before this, and
```
nex3, line 208: It doesn't really make sense to generalize this list to other functions without also modifying the rest of the paragraph. Otherwise mentioning `round()` and `abs()` here feels like a non sequitur, since they don't have anything to do with the previous sentence.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +161,22 @@ CSS to unitless numbers:
...
 
-[`min()` and `max()` functions]: https://drafts.csswg.org/css-values-4/#calc-notation
+[calculation functions]: https://drafts.csswg.org/css-values-4/#calc-notation
 [to support the iPhoneX]: https://webkit.org/blog/7929/designing-websites-for-iphone-x/
-[`min()`]: ../modules/math#min
-[`max()`]: ../modules/math#max
+[math functions]: ../modules/math
 
-If a `min()` or `max()` function call is a valid calculation expression, it will
+If a calculation function call is a valid calculation expression, it will
```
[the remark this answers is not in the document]
nex3: I think this is still an issue. Not every calculation function call is parsed as a Sass math function.

source/documentation/values/calculations.html.md.erb
```diff
@@ -161,42 +171,55 @@ CSS to unitless numbers:
...
+[`abs()`]: ../modules/math#abs
 [`min()`]: ../modules/math#min
 [`max()`]: ../modules/math#max
 
-If a `min()` or `max()` function call is a valid calculation expression, it will
+If a calculation function call is a valid calculation expression, it will
 be parsed as a calculation. But as soon as any part of the call contains a
 SassScript feature that isn't supported in a calculation, like the [modulo
-operator], it's parsed as a call to Sass's core `min()` or `max()` function
+operator], it's parsed as a call to the appropriate Sass math
```
nex3, line 222:
```suggestion
operator], it's parsed as a call to the appropriate Sass math function
```

Also, please reflow paragraphs once you edit them. (Most editors have some built-in or plugin command for this.)

source/documentation/values/calculations.html.md.erb
```diff
@@ -210,12 +233,34 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
   from, or compared to numbers with units. `min()` and `max()` are different,
   though: for backwards-compatibility with the global Sass `min()` and `max()`
   functions which allow unit/unitless mixing for historical reasons, these units
-  can be mixed as long as they're contained directly within a `min()` or `max()`
+  can be mixed as long as they're contained directly within a `min()` or `max()`
```
nex3, line 236: Nit: trailing whitespace

source/documentation/values/calculations.html.md.erb
```diff
@@ -210,12 +233,34 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
   from, or compared to numbers with units. `min()` and `max()` are different,
   though: for backwards-compatibility with the global Sass `min()` and `max()`
   functions which allow unit/unitless mixing for historical reasons, these units
-  can be mixed as long as they're contained directly within a `min()` or `max()`
+  can be mixed as long as they're contained directly within a `min()` or `max()` 
   calculation.
+  For instance, `sin($number: 4)` will throw an error, as `sin()` was 
+  never a global Sass function.
```
nex3, line 239: I think this example is a bit misleading. It doesn't involve a unitless number being added to, subtracted from, compared to a number with units. Instead, it shows a unitless number being passed to `sin()` by name, neither of which are the subject of this `heads_up`.

source/documentation/values/calculations.html.md.erb
```diff
@@ -210,12 +233,34 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
 <% end %>
```
nex3, line 240: Since this `heads_up` is specific to `min` and `max`, it should probably be under that heading (if you end up keeping it).

source/documentation/values/calculations.html.md.erb
```diff
@@ -210,12 +233,34 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
 
 <!-- TODO(nweiz): auto-generate this CSS once we're compiling with Dart Sass -->
 
+### [`min()`] and [`max()`]
```
nex3, line 244: As a rule of thumb, a section of documentation should always have some amount of prose, in addition to any callouts or examples it contains. Otherwise, it's a sign that that heading isn't actually a useful division, and its contents should be part of the parent section instead.

nex3, line 244: Style nit: A heading should always be surrounded by empty lines.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
     padding-right: 20px;
   }
 <% end %>
+
+### [`round()`]
+<% impl_status dart: "1.64.0", libsass: false, ruby: false, feature: "Special function syntax" do %>
+  LibSass, Ruby Sass, and versions of Dart Sass prior to 1.64.0 *always* parse
+  `round()` a as Sass function. To create a plain CSS calculation function
```
nex3, line 310:
```suggestion
  `round()` as a Sass function. To create a plain CSS calculation function
```

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+  call for those implementations, you can write something like
```
nex3, line 311: "plain CSS calculation function call" is a little confusing here, since a "function call" and a "calculation" are usually different things in CSS. I'd probably just say "plain CSS calculation".

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+  `round(#{$number})` instead.
```
nex3, line 312: This example is misleading, since `round()` with a single number argument isn't valid CSS.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+
+  [special functions]: ../syntax/special-functions
+<% end %>
+
+The `round(<strategy>, value, step)` function accepts an optionl rounding strategy,
```
nex3, line 317:
```suggestion
The `round(<strategy>, value, step)` function accepts an optional rounding strategy, 
```

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+  `round()` a as Sass function. To create a plain CSS calculation function
+  call for those implementations, you can write something like
+  `round(#{$number})` instead.
+
+  [special functions]: ../syntax/special-functions
+<% end %>
+
+The `round(<strategy>, value, step)` function accepts an optionl rounding strategy, 
+a value to be rounded and a rounding interval `step`. 
+`strategy` should be `nearest`, `up`, `down`, or `to-zero`.
```
nex3, line 319: Currently, this doesn't explain what `strategy` or `step` _mean_. You don't necessarily need to document that fully here, but you should at least link to the MDN documentation for the CSS function.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+`round(value)` is also supported for backwards compatibility.
```
nex3, line 320: It's probably a good idea to encourage users to supply a `step` parameter rather than relying on this backwards-compatibility.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+    padding-right: round($number + 10px)
+    padding-bottom: round($number + 10px, $step + 10%)
+
+  ===
+  .post {
+    padding-left: 15px;
+    padding-right: 23px;
+    padding-bottom: round(22.5px, 15px + 10%);
+  }
+<% end %>
```
nex3, line 351: I don't know that each legacy function necessarily needs its own separate example, since the crucial point—when they're parsed as calculations versus Sass function calls—is the same for all of them. It might be better to just edit the `min`/`max` example to contain a couple examples of other functions instead of exclusively using `min` and `max`.

source/documentation/values/calculations.html.md.erb
```diff
@@ -258,3 +303,81 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
+  `abs()` a as Sass function. To create a plain CSS calculation function
+  call for those implementations, you can write something like
+  `abs(#{$number})` instead.
+
+  [special functions]: ../syntax/special-functions
+<% end %>
+
+Given that CSS resolves percentages before calculations, 
+the global `abs(value)` function compatibiliy with % unit parameters is deprecated.
+In the future, this will emit a CSS abs() function to be resolved by the browser.
```
nex3, line 365: This is definitely worth calling out, although it might be better in a `heads_up` block because it's a warning about a potential trap users could fall into. I find this phrasing pretty confusing, though: it's not immediately clear what "CSS resolves percentages before calculations" means in practice to stylesheet authors, and "compatibility with % unit parameters" is pretty abstract. Consider instead starting from a concrete example, explaining how it works in CSS, then explaining how it works in Sass and why it has to be different.

nex3 (changes requested):

source/documentation/values/calculations.html.md.erb
pamelalozano16: Done!

source/documentation/values/calculations.html.md.erb
pamelalozano16: Hm maybe I should just delete this sentence to not encourage users to use a single argument round? Since the rest of the paragraph mentions a required value and step.

source/documentation/values/calculations.html.md.erb
pamelalozano16: Got it, done!

source/documentation/values/calculations.html.md.erb
pamelalozano16: Got it, I added a sentence after the title, thanks! :)

PR #755, opened 2023-06-30
Update calculation functions documentation

pamelalozano16:
New PR because of transition to NodeJS. Closed PR: https://github.com/sass/sass-site/pull/738

### DIFF
--- a/source/documentation/breaking-changes/abs-percent.md
+++ b/source/documentation/breaking-changes/abs-percent.md
@@ -1,0 +1,21 @@
+---
+title: 'Breaking Change: abs() Percentage'
+introduction: >
+  Sass has historically supported the `abs()` function. After CSS supported
+  calculations in Values and Units Level 4, we had to workaround
+  backwards-compatibility. However, for the `abs()` function we posses a
+  compatibility problem supporting the `%` unit.
+---
+
+The `abs()` global function in Sass supported the `%` unit as an input and would
+resolve the `abs()` function before resolving the `%` value. For instance, if
+the input was `abs(10%)` the function will return `10%`. As a result, if the
+value of `10%` represented `-50px` the function would return `-50px`.
+
+However, the CSS `abs()` abs function will resolve the `%` before resolving the
+function. Therefore if the value of `10%` represented `-50px`, `abs(10%)` would
+return `-10%` which in the browser would be `50px`.
+
+For this reason, we are deprecating the global abs() function with a percentage.
+To preserve the current behavior, use `math.abs()` or `abs(#{})` instead.
+
--- a/source/documentation/breaking-changes/abs-percent.md
+++ b/source/documentation/breaking-changes/abs-percent.md
@@ -17,5 +17,5 @@ function. Therefore if the value of `10%` represented `-50px`, `abs(10%)` would
 return `-10%` which in the browser would be `50px`.
 
 For this reason, we are deprecating the global abs() function with a percentage.
-To preserve the current behavior, use `math.abs()` or `abs(#{})` instead.
+To preserve the current behavior, use `math.abs()` instead.
 
--- a/source/documentation/values/calculations.md
+++ b/source/documentation/values/calculations.md
@@ -6,7 +6,7 @@ introduction: >
   much as possible, even if they're combined with one another.
 ---
 
-{% compatibility 'dart: "1.40.0"','libsass: false', 'ruby: false'%}
+{% compatibility 'dart: "1.40.0"', 'libsass: false', 'ruby: false' %}
   LibSass, Ruby Sass, and versions of Dart Sass prior to 1.40.0 parse `calc()`
   as a [special function] like `element()`.
 
@@ -158,24 +158,27 @@ CSS to unitless numbers:
 
 ## Calculation Functions
 
-{% compatibility 'dart: "1.64.0"', 'libsass: false', 'ruby: false', 'feature: "Special function syntax"' %}
-  LibSass, Ruby Sass, and versions of Dart Sass subsequent to 1.64.0 support the
-  following calculation functions, with the exception of `min()`, `max()`, and
-  `clamp()`.
+{% compatibility 'dart: "1.65.0"', 'libsass: false', 'ruby: false', 'feature: "Additional functions"' %}
+  Versions of Dart Sass subsequent to 1.65.0 handle the execution of these
+  calculation functions: `round()`, `mod()`, `rem()`, `sin()`, `cos()`, `tan()`,
+  `asin()`, `acos()`, `atan()`, `atan2()`, `pow()`, `sqrt()`, `hypot()`,
+  `log()`, `exp()`, `abs()`, and `sign()`.
 {% endcompatibility %}
 
 Sass parses the following functions as [calculations]:
 * Comparison Functions: [`min()`], [`max()`], and [`clamp()`]
 * Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
-* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`], [`atan()`], and [`atan2()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`],
+  [`atan()`], and [`atan2()`].
 * Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
 * Sign-Related Functions: [`abs()`] and [`sign()`].
 
 [calculations]: https://www.w3.org/TR/css-values-4/#math
-[`min()`]: #min-and-max
-[`max()`]: #min-and-max
-[`clamp()`]: /documentation/modules/math/#clamp
-[`round()`]: #round
+[`min()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/min
+[`max()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/max
+[`clamp()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/clamp
+[`round()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/round
+[`abs()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/abs
 [`sin()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sin
 [`cos()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/cos
 [`tan()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/tan
@@ -187,7 +190,6 @@ Sass parses the following functions as [calculations]:
 [`sqrt()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sqrt
 [`hypot()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/hypot
 [`log()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/log
-[`abs()`]: #abs
 [`exp()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/exp
 [`mod()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/mod
 [`rem()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/rem
@@ -201,18 +203,17 @@ CSS added support for [mathematical expressions] in Values and Units Level
 those existing stylesheets. This led to the need for extra-special syntactic
 cleverness.
 
-[`min()` and `max()` functions]: https://www.w3.org/TR/css-values-4/#math
-[to support the iPhoneX]: https://webkit.org/blog/7929/designing-websites-for-iphone-x/
+[mathematical expressions]: https://www.w3.org/TR/css-values-4/#math
 [`round()`]: ../modules/math#round
 [`abs()`]: ../modules/math#abs
 [`min()`]: ../modules/math#min
 [`max()`]: ../modules/math#max
 
-If a calculation function call is a valid calculation expression, it will
-be parsed as a calculation. But as soon as any part of the call contains a
-SassScript feature that isn't supported in a calculation, like the [modulo
-operator], it's parsed as a call to the appropriate Sass math function
-instead.
+If a call to `round()`, `abs()`, `min()`, or `max()` is a valid calculation
+expression, it will be parsed as a calculation. But as soon as any part of the
+call contains a SassScript feature that isn't supported in a calculation, like
+the [modulo operator], it's parsed as a call to the appropriate Sass math
+function instead.
 
 Since calculations are simplified to numbers when possible anyway, the only
 substantive difference is that the Sass functions only support units that can be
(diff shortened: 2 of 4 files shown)


source/documentation/breaking-changes/abs-percent.md
```diff
@@ -0,0 +1,21 @@
...
+resolve the `abs()` function before resolving the `%` value. For instance, if
+the input was `abs(10%)` the function will return `10%`. As a result, if the
+value of `10%` represented `-50px` the function would return `-50px`.
+
+However, the CSS `abs()` abs function will resolve the `%` before resolving the
+function. Therefore if the value of `10%` represented `-50px`, `abs(10%)` would
+return `-10%` which in the browser would be `50px`.
+
+For this reason, we are deprecating the global abs() function with a percentage.
+To preserve the current behavior, use `math.abs()` or `abs(#{})` instead.
```
nex3, line 20: I don't think we should suggest `abs(#{})`, since that returns either a string or a calculation object, and the current behavior returns a number.

source/documentation/values/calculations.md
```diff
@@ -6,7 +6,7 @@ introduction: >
   much as possible, even if they're combined with one another.
 ---
 
-{% compatibility 'dart: "1.40.0"', 'libsass: false', 'ruby: false' %}
+{% compatibility 'dart: "1.40.0"','libsass: false', 'ruby: false'%}
```
nex3, line 9: Accidentally deleted spaces?

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
-  other than interpolation, like variables or function calls.
-
-  Dart Sass 1.41.0 parses `min()` and `max()` functions as calculations, but
-  doesn't allow unitless numbers to be combined with numbers with units. This
-  was backwards-incompatible with the global `min()` and `max()` functions, so
-  that behavior was reverted.
-
-  [special functions]: /documentation/syntax/special-functions
+{% compatibility 'dart: "1.64.0"', 'libsass: false', 'ruby: false', 'feature: "Special function syntax"' %}
+  LibSass, Ruby Sass, and versions of Dart Sass subsequent to 1.64.0 support the
```
nex3, line 162: Note that 1.64.0 has already been released, so this version number will have to be updated to whatever version actually ends up having this feature.

nex3, line 162: LibSass and Ruby Sass don't support _any_ calculation functions.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
+  following calculation functions, with the exception of `min()`, `max()`, and
+  `clamp()`.
```
nex3, line 164: Two things:

1. It's better to list the functions that _are_ supported in the given version than to just say "everything but X Y and Z". That's clearer to the reader as well as less likely to become incorrect if new functions are added later.

2. "support the following calculation" functions is vague. The point of these compatibility notes is to give the reader a clear understanding of how behavior differs across versions and implementations, and just saying that something isn't supported doesn't clarify what actually happens if you try to use it. That's why most of these notes are framed as explaining how it worked _before_ the new feature was added.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
 [to support the iPhoneX]: https://webkit.org/blog/7929/designing-websites-for-iphone-x/
-[`min()`]: /documentation/modules/math#min
-[`max()`]: /documentation/modules/math#max
+[`round()`]: ../modules/math#round
+[`abs()`]: ../modules/math#abs
+[`min()`]: ../modules/math#min
+[`max()`]: ../modules/math#max
 
-If a `min()` or `max()` function call is a valid calculation expression, it will
+If a calculation function call is a valid calculation expression, it will
```
nex3, line 211: I think it's worth specifically saying "a call to `round()`, `abs()`, `min()`, or `max()`" here, since this isn't true for _all_ calculations.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
 
 Since calculations are simplified to numbers when possible anyway, the only
 substantive difference is that the Sass functions only support units that can be
 combined at build time, so `min(12px % 10, 10%)` will throw an error.
 
-[modulo operator]: /documentation/operators/numeric
+[modulo operator]: /documentation/operators/numeric/
+
+#### `min()` and `max()`
+{% compatibility 'dart: ">=1.11.0 <1.42.0"', 'libsass: false', 'ruby: false', 'feature: "min and max syntax"' %}
```
nex3, line 224: Nit: Newline above this

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
+  was backwards-incompatible with the global `min()` and `max()` functions, so
+  that behavior was reverted.
+
+  [special functions]: /documentation/syntax/special-functions/
+{% endcompatibility %}
 
 {% headsUp %}
   Other calculations don't allow unitless numbers to be added to, subtracted
-  from, or compared to numbers with units. `min()` and `max()` are different,
+  from, or compared to numbers with units. [`min()`] and [`max()`] are different,
```
nex3, line 251: These links don't currently work, because callouts are resolved in a different Markdown context than the main page.

source/documentation/values/calculations.md
```diff
@@ -225,8 +283,23 @@ combined at build time, so `min(12px % 10, 10%)` will throw an error.
...
     padding-right: max($padding % 10, 20px);
   }
+
+  .post-image {
+    // Since these round() calls are valid calculation expressions, they're
+    // parsed as calculations.
+    padding-left: round(nearest, $number, $step);
+    padding-right: round($number + 10px);
+    padding-bottom: round($number + 10px, $step + 10%);
+  }
```
nex3, line 293: These examples should be under the headings for their respective functions, not under `min()` and `max()`.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
-  plain CSS, but parse them as Sass functions if they contain Sass features
-  other than interpolation, like variables or function calls.
-
-  Dart Sass 1.41.0 parses `min()` and `max()` functions as calculations, but
-  doesn't allow unitless numbers to be combined with numbers with units. This
-  was backwards-incompatible with the global `min()` and `max()` functions, so
-  that behavior was reverted.
-
-  [special functions]: /documentation/syntax/special-functions
+{% compatibility 'dart: "1.64.0"', 'libsass: false', 'ruby: false', 'feature: "Special function syntax"' %}
```
nex3, line 161: The feature tag "special function syntax" is misleading, since we usually use that term to refer to [functions that are parsed as strings](https://sass-lang.com/documentation/syntax/special-functions/). Really what this is describing is the version all the new functions got added in, so maybe "additional functions" is a better name.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
-Sass supported its own [`min()`] and [`max()`] functions long before this, and
-it needed to be backwards-compatible with all those existing stylesheets. This
-led to the need for extra-special syntactic cleverness.
-
-[`min()` and `max()` functions]: https://drafts.csswg.org/css-values-4/#calc-notation
+Sass parses the following functions as [calculations]:
+* Comparison Functions: [`min()`], [`max()`], and [`clamp()`]
+* Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`], [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
```
nex3, line 171: Wrap lines at 80 characters

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
   can be mixed as long as they're contained directly within a `min()` or `max()`
   calculation.
+
+  For instance, `min(5 + 10px, 20px)` will result in `15px`. However
+  `sqrt(5 + 10px)` will throw an error, as `sqrt(5 + 10px)` was never a global
+  Sass function, and these are incompatible units.
 {% endheadsUp %}
 
+[`min()`] and [`max()`] functions take one or more comma-separated expressions
+as parameters.
```
nex3, line 263: I don't think you need to document these functions' behavior or parameters; that's already been explained in the MDN links.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
+{% endcompatibility %}
 
 {% headsUp %}
   Other calculations don't allow unitless numbers to be added to, subtracted
-  from, or compared to numbers with units. `min()` and `max()` are different,
+  from, or compared to numbers with units. [`min()`] and [`max()`] are different,
   though: for backwards-compatibility with the global Sass `min()` and `max()`
   functions which allow unit/unitless mixing for historical reasons, these units
   can be mixed as long as they're contained directly within a `min()` or `max()`
   calculation.
```
nex3, line 255: This logic applies to `abs()` and single-argument `round()` as well, so we should probably move it up to the parent section.

nex3 (changes requested):

source/documentation/values/calculations.md
pamelalozano16: I updated them to the [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/CSS/min) instead

source/documentation/values/calculations.md
pamelalozano16: In a [previous comment](https://github.com/sass/sass-site/pull/738#discussion_r1236009723) you mentioned we should have all the examples together, should I change it back to how it was?

source/documentation/values/calculations.md
pamelalozano16: Thanks! I listed the functions and changed the word `support` for `handle the execution of` which I think is more accurate.

source/documentation/values/calculations.md
nex3: Ah, sorry for the inconsistency. I think there are two possible ways to go here:

1. Have thorough examples with multiple properties, like you do here, for each function.
2. Have a single style rule (such as `.post-image`) that uses each function _once_.

The way it is now is strange because the example is already visually separated into several distinct blocks, but those blocks don't correspond to the overarching structure of the page.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
+* Comparison Functions: [`min()`], [`max()`], and [`clamp()`]
+* Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`],
+  [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
+* Sign-Related Functions: [`abs()`] and [`sign()`].
+
+[calculations]: https://www.w3.org/TR/css-values-4/#math
+[`min()`]: #min-and-max
+[`max()`]: #min-and-max
```
nex3, line 178: These should link to the MDN docs, since the documentation below doesn't fully explain the calculation semantics.

source/documentation/values/calculations.md
```diff
@@ -156,61 +156,119 @@ CSS to unitless numbers:
...
 substantive difference is that the Sass functions only support units that can be
 combined at build time, so `min(12px % 10, 10%)` will throw an error.
 
-[modulo operator]: /documentation/operators/numeric
+[modulo operator]: /documentation/operators/numeric/
 
 {% headsUp %}
   Other calculations don't allow unitless numbers to be added to, subtracted
-  from, or compared to numbers with units. `min()` and `max()` are different,
+  from, or compared to numbers with units. [`min()`] and [`max()`] are different,
```
nex3, line 227: This should list `abs()` and single-argument `round()` as well.

nex3 (changes requested):

source/documentation/values/calculations.md
pamelalozano16: I think the first option would be better, just to provide more examples. :) Lmk your thoughts

source/documentation/values/calculations.md
```diff
@@ -156,59 +156,117 @@ CSS to unitless numbers:
...
 combined at build time, so `min(12px % 10, 10%)` will throw an error.
 
-[modulo operator]: /documentation/operators/numeric
+[modulo operator]: /documentation/operators/numeric/
 
 {% headsUp %}
   Other calculations don't allow unitless numbers to be added to, subtracted
-  from, or compared to numbers with units. `min()` and `max()` are different,
-  though: for backwards-compatibility with the global Sass `min()` and `max()`
+  from, or compared to numbers with units. [`round()`], [`abs()`], [`min()`] and [`max()`]
```
nex3, line 227: Style nit: wrap lines at 80 characters

nex3, line 227: Make it clear that this only applies to _single-argument_ `round()` functions.

source/documentation/values/calculations.md
```diff
@@ -156,59 +156,117 @@ CSS to unitless numbers:
...
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`],
+  [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
+* Sign-Related Functions: [`abs()`] and [`sign()`].
+
+[calculations]: https://www.w3.org/TR/css-values-4/#math
+[`min()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/min
+[`max()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/max
+[`clamp()`]: /documentation/modules/math/#clamp
+[`round()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/round
```
pamelalozano16, line 180: Even though I explain these functions in a following section, I thought it'd be better to link this ones to mdn here as well.

source/documentation/values/calculations.md
```diff
@@ -156,59 +156,117 @@ CSS to unitless numbers:
...
+* Stepped Value Functions: [`round()`], [`mod()`], and [`rem()`].
+* Trigonometric Functions: [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`],
+  [`atan()`], and [`atan2()`].
+* Exponential Functions: [`pow()`], [`sqrt()`], [`hypot()`], [`log()`], and [`exp()`].
+* Sign-Related Functions: [`abs()`] and [`sign()`].
+
+[calculations]: https://www.w3.org/TR/css-values-4/#math
+[`min()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/min
+[`max()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/max
+[`clamp()`]: /documentation/modules/math/#clamp
```
nex3, line 179: `clamp()` should also link to MDN

nex3 (approved):

source/documentation/values/calculations.md
pamelalozano16: This is wrapped I think 🤔

source/documentation/values/calculations.md
pamelalozano16: Would this be `1.65.0`? 🤔

Goodwine (approved):

PR #768, opened 2023-07-08
Fix diff link comment

jathak:
The comment won't actually work until this is merged, since the secondary workflow is still running based on main, but I believe this should fix the issue with the dependabot PRs

### DIFF
--- a/.github/workflows/shadow-repo-update.yml
+++ b/.github/workflows/shadow-repo-update.yml
@@ -45,27 +45,49 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'workflow_run'
     steps:
-      - name: Clone shadow repo
-        uses: actions/checkout@v3
-        with:
-          repository: ${{ vars.SHADOW_OWNER }}/${{ vars.SHADOW_REPO }}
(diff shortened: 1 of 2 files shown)


.github/workflows/shadow-repo-update.yml
```diff
@@ -69,3 +69,21 @@ jobs:
           author_email: sass.bot.beep.boop@gmail.com
           message: Update from https://github.com/${{ github.repository }}/commit/${{ github.sha }}
           commit: --allow-empty
+
+  diff-link:
```
nex3, line 73: If you make this run after the push command, you can get rid of the "(may take a few minutes)" text since the comment will only post once the diff is actually visible.

nex3 (changes requested):

.github/workflows/shadow-repo-update.yml
jathak: I was originally thinking that, but the push runs each time the branch is pushed to while the comment is only made when the PR is open, so I don't think they can be linked together that way

nex3 (approved):

PR #770, opened 2023-07-10
Add OddBird team members

jerivas:
Sondra and Sana also worked on the playground and 11ty port, just didn't interact much with the Sass team


nex3:
Thanks for catching this!

PR #792, opened 2023-08-06
fix(docs): correct default import example about default export

nazarepiedady:
I am applying a subtle typo correction to the `breaking-changes/default-export.md` document in this pull request.
I found this typo while translating this document for the translated version of the website I am working on.

@nex3

### DIFF
--- a/source/documentation/breaking-changes/default-export.md
+++ b/source/documentation/breaking-changes/default-export.md
@@ -35,4 +35,4 @@ import * as sass from 'sass'; // Do this
 Until Dart Sass 2.0.0, we will continue to support users loading Sass's default
 export. The first time any properties on the default export are accessed, it
 will emit a deprecation warning to `console.error()`. To avoid this error, use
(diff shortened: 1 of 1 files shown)


nazarepiedady:
@nex3 is this issue already fixed?
Because I could close this pull request if so.

nex3:
I don't think this is fixed, but it does look like you need to rebase this onto the latest main branch.

nazarepiedady:
@nex3 the pull request is updated, I believe that now is possible to go ahead to the next stage.

nex3:
This still isn't rebased against the latest HEAD (you can tell because it has a bunch of duplicate commits listed, instead of just one small commit).

nazarepiedady:
@nex3 I have no idea, and I cannot see a conflict until now. If it is not possible to do the merge, could you use the `cherry-pick` to only add to the main change I have made?

Goodwine (approved):

PR #798, opened 2023-08-10
Calc functions breaking change page

### LINKED ISSUE #797: Update @function documentation for css calc function changes
Since 1.65.0 `pow()` is always parsed as calculation instead of custom function. If user defines a custom function with the same name as native css function, Sass would not parse it as function any more.

It means the example that defines a custom `pow()` function is no longer correct and need to be updated: https://sass-lang.com/documentation/at-rules/function/

```
@function pow($base, $exponent) {
  $result: 1;
  @for $_ from 1 through $exponent {
    $result: $result * $base;
  }
  @return $result;
}

// This example still produce the same output, but the custom function is not called due to `pow(4, 3)` being parsed as calc.
.sidebar {
  float: left;
  margin-left: pow(4, 3) * 1px;
}
```

Also, it is probably worth to have a breaking change notice as some users got bite by this behavior change:
- https://github.com/sass/dart-sass/issues/2059
- https://github.com/sass/dart-sass/issues/2060

@pamelalozano16

pamelalozano16:
Fix #797.

### DIFF
--- a/source/documentation/at-rules/function.md
+++ b/source/documentation/at-rules/function.md
@@ -18,30 +18,30 @@ Functions are called using the normal CSS function syntax.
 [`@return` at-rule]: #return
 
 {% codeExample 'functions' %}
-  @function pow($base, $exponent) {
-    $result: 1;
-    @for $_ from 1 through $exponent {
-      $result: $result * $base;
+  @function fibonacci($n) {
+    $sequence: 0 1;
+    @for $_ from 1 through $n {
+      $new: nth($sequence, length($sequence)) + nth($sequence, length($sequence) - 1);
+      $sequence: append($sequence, $new);
     }
-    @return $result;
+    @return nth($sequence, length($sequence));
   }
 
   .sidebar {
     float: left;
-    margin-left: pow(4, 3) * 1px;
+    margin-left: fibonacci(4) * 1px;
   }
   ===
-  @function pow($base, $exponent)
-    $result: 1
-    @for $_ from 1 through $exponent
-      $result: $result * $base
-
-    @return $result
-
+  @function fibonacci($n)
+    $sequence: 0 1
+    @for $_ from 1 through $n
+      $new: nth($sequence, length($sequence)) + nth($sequence, length($sequence) - 1)
+      $sequence: append($sequence, $new)
+    @return nth($sequence, length($sequence))
 
   .sidebar
     float: left
-    margin-left: pow(4, 3) * 1px
+    margin-left: fibonacci(4) * 1px
 {% endcodeExample %}
 
 {% funFact %}
--- a/source/documentation/breaking-changes/calc-functions.md
+++ b/source/documentation/breaking-changes/calc-functions.md
@@ -1,0 +1,123 @@
+---
+title: 'Breaking Change: Calc functions'
+introduction: >
+   Sass has added support for the CSS Values and Units Level 4 specification.
+   These global functions are now reserved: `round()`, `mod()`, `rem()`,
+   `sin()`, `cos()`, `tan()`, `asin()`, `acos()`, `atan()`, `atan2()`, `pow()`,
+   `sqrt()`, `hypot()`, `log()`, `exp()`, `abs()`, and `sign()`.
+-------------------------------------------------------------------------------
+{% compatibility 'dart: "1.65.0"', 'libsass: false', 'ruby: false' %}{% endcompatibility %}
+
+The new calc functions are now reserved as global functions. These function
+calls will be parsed as CSS calculations instead of function calls: [`round()`], [`mod()`],
+[`rem()`], [`sin()`], [`cos()`], [`tan()`], [`asin()`], [`acos()`], [`atan()`], [`atan2()`],
+[`pow()`], [`sqrt()`], [`hypot()`], [`log()`], [`exp()`], [`abs()`], or [`sign()`].
+User-defined functions named after any of these may no longer be called without
+a namespace.
+
+[`round()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/round
+[`abs()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/abs
+[`sin()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sin
+[`cos()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/cos
+[`tan()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/tan
+[`asin()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/asin
+[`acos()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/acos
+[`atan()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/atan
+[`atan2()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/atan2
+[`pow()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/pow
+[`sqrt()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sqrt
+[`hypot()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/hypot
+[`log()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/log
+[`exp()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/exp
+[`mod()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/mod
+[`rem()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/rem
+[`sign()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sign
+
+{% codeExample 'calc-function-bad', false %}
+@function rem($number) {
+  @return $number % 10;
+}
+
+.sidebar {
+  // WRONG, Sass parses `rem()` as a calculation, not as a call to the function
+  // declared above. This throws an error because `rem()` requires two parameters.
+  margin-left: rem(9) * 1px;
+}
+===
+@function rem($number)
+  @return $number % 10
+
+.sidebar 
+    // WRONG, Sass parses `rem()` as a calculation, not as a call to the function
+    // declared above. This throws an error because `rem()` requires two parameters.
+    margin-left: rem(9) * 1px
+{% endcodeExample %}
+
+However, if the function is loaded from another file and called with a namespace
+it is parsed as a function call.
+
+{% codeExample 'calc-function-use' %}
+  // _library.scss
+  @function rem($number) {
+    @return $number % 10;
+  }
+  ---
+  // style.scss
+  @use 'library';
+
+  .sidebar {
+    // RIGHT, Sass parses `library.rem()` as a function call rather than
+    // a calculation because of the namespace.
+    margin: library.rem(11);
+  }
+  ===
+  // _library.sass
+  @function rem($number)
+    @return $number % 10
+  ---
+  // style.sass
+  @use 'library'
+
+  .sidebar
+    // RIGHT, Sass parses `library.rem()` as a function call rather than
+    // a calculation because of the namespace.
+    a: library.rem(11) * 1px
+  ===
+  .sidebar {
+    margin: 1;
+  }
+{% endcodeExample %}
+
+{% headsUp %}
+  Because calling a function from a module loaded [_without a namespace_] by
+  writing `@use <url> as *` has the same syntax as calling a global function,
+  Sass will parse these function calls as calculations, not as calls to the
+  function defined in the module!
+
+  [_without a namespace_]: /documentation/at-rules/use/#choosing-a-namespace
+{% endheadsUp %}
+
+To preserve the current behavior, use [`meta.call()`] and [`meta.get-function()`].
+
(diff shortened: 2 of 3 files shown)


source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+[`sign()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/sign
+
+{% codeExample 'calc-function-bad', false %}
+@function rem($number) {
+  @return $number % 10;
+}
+
+/* This example would parse `rem()` as a calc function, not as the declared
+function above. Since `rem()` requires two parameters, this would throw a
+parameter error. */
```
Goodwine, line 41: Use silent comments (`//`) instead of loud comments `/* */`

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+.sidebar {
+  margin-left: rem(9) * 1px;
```
Goodwine, line 43: It would make more sense to move the comment above this line instead of over the selector

source/documentation/breaking-changes/calc-functions.md
Goodwine, line 41: You can be more active with the wording of this comment, and following other examples in other breaking change announcements you can add "WRONG" at the beginning to indicate it will cause an error
```
// WRONG, Sass parses `rem()` as a calculation, not as a call to the function
// declared above. This throws an error because `rem()` requires two parameters.
```

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+===
+@function rem($number)
+  @return $number % 10
+
+// This example would parse `rem()` as a calc function, not as the declared
+// function above. Since `rem()` requires two parameters, this would throw a
+// parameter error.
+.sidebar 
+  margin-left: rem(9) * 1px
+
```
Goodwine, line 55: nit: remove this line

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+{% endcodeExample %}
+
+However, if the custom `rem()` function was imported to another file, it would
+be parsed as the declared function since it's assigned to a different namespace.
```
Goodwine, line 59: Use "loaded" instead of "imported" to avoid confusion with the `@import` at-rule, also since this is outside an example mentioning `rem()` makes it seem like this is rem-specific when it affects all others too
```suggestion
However, if the function is loaded from another file and called with a namespace
it is parsed as a function call.
```

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+
+{% codeExample 'calc-function-import', false %}
```
Goodwine, line 61: You can drop `, false` from here since this is supposed to be a working example

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+<===> foo.scss
```
Goodwine, line 62: This should use `//` instead of `<===>`. Also per other examples using `// _library.scss` seems to be better in this case

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+@function rem($number) {
+  @return $number % 10;
+}
+
+<===> main.scss
```
Goodwine, line 67: `codeExample` splits examples with `---`, see: https://github.com/sass/sass-site/blob/495d514a215712089ad1de89c1533c7f534260ee/source/_includes/code_snippets/example-use-with.liquid#L11

```suggestion
---
// style.scss
```

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+// This example would not throw an error.
+@use 'foo';
```
Goodwine, line 69: if you rename the file to `_library.scss` make sure to update the `@use` here and the references below

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+be parsed as the declared function since it's assigned to a different namespace.
+
+{% codeExample 'calc-function-import', false %}
+<===> foo.scss
+@function rem($number) {
+  @return $number % 10;
+}
+
+<===> main.scss
+// This example would not throw an error.
```
Goodwine, line 68: I think this comment is redundant with the explanation from the paragraph above. But if you decide to keep it make sure to explain "why", for example..
```
// RIGHT, Sass parses `library.rem()` as a function call rather than
// a calculation because of the namespace.
```

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+@use 'foo';
+
+.sidebar {
+  a: foo.rem(9) * 1px;
+}
+===
+<===> foo.scss
```
Goodwine, line 75: Make sure to update the `.sass` variant and make sure the file names end with `.sass` rather than `.scss`

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+  If the module was imported [_without a namespace_] 
+  calc functions would override the custom declared functions.
+  For instance, `@use 'foo' as *` would change the call to `a: rem(10)`,
+  which would result in an error.
+
+[_without a namespace_]: /documentation/at-rules/use/#choosing-a-namespace
+{% endheadsUp %}
+
+To preserve the current behavior, use the [`call()`] and [`get-function()`] methods
+from the `meta` module.
```
Goodwine, line 98: In code samples, it's more common to see a function prefixed with its module. Replace `call()` with `meta.call()`. Same with `get-function()`. You can then remove the mention of "from the meta module"

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+
+[`get-function()`]:http://localhost:8080/documentation/modules/meta/#get-function
+[`call()`]: /documentation/modules/meta/#call
```
Goodwine, line 101: 1. make sure to update these per the comment above
2. don't use the localhost:8080 url

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+{% codeExample 'function-using-meta', false %}
```
Goodwine, line 102: Nit: add a separation line above this line

Goodwine, line 102: You can remove `false` from this codeExample since it's supposed to be a working example

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+.sidebar
+  a: foo.rem(9) * 1px
+
+{% endcodeExample %}
+
+{% headsUp %}
+  If the module was imported [_without a namespace_] 
+  calc functions would override the custom declared functions.
+  For instance, `@use 'foo' as *` would change the call to `a: rem(10)`,
+  which would result in an error.
```
Goodwine, line 92:
```suggestion
  Because calling a function from a module loaded [_without a namespace_]
  by writing `@use <url> as *` has the same syntax as calling a global function,
  Sass will parse these function calls as calculations, not as calls to the
  function defined in the module!
```

Goodwine (changes requested):

ntkme:
We should also update the following code example using `pow()` to something else on https://sass-lang.com/documentation/at-rules/function/ 

https://github.com/sass/sass-site/blob/495d514a215712089ad1de89c1533c7f534260ee/source/documentation/at-rules/function.md?plain=1#L20-L45

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,121 @@
...
+  functions: `round()`, `mod()`, `rem()`, `sin()`, `cos()`, `tan()`, `asin()`,
+  `acos()`, `atan()`, `atan2()`, `pow()`, `sqrt()`, `hypot()`, `log()`, `exp()`,
+  `abs()`, and `sign()`.
+--------------------------------------------------------------------------------
+
+These calc functions are now reserved words, which means any custom defined
+functions defined with: [`round()`], [`mod()`], [`rem()`], [`sin()`], [`cos()`], [`tan()`],
+[`asin()`], [`acos()`], [`atan()`], [`atan2()`], [`pow()`], [`sqrt()`], [`hypot()`],
+[`log()`], [`exp()`], [`abs()`], or [`sign()`] will be parsed as calculations
+instead of functions.
```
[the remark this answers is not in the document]
pamelalozano16: Great suggestion!

source/documentation/breaking-changes/calc-functions.md
pamelalozano16: This causes trouble in the 'Sass' example for some reason 🤔 
```
.sidebar 
// WRONG, Sass parses `rem()` as a calculation, not as a call to the function
// declared above. This throws an error because `rem()` requires two parameters.
    margin-left: rem(9) * 1px
```
This looks like [this](https://screenshot.googleplex.com/4Q7p3tWU5ANwyTH). For this reason I'll leave the comment before the selector.

Goodwine: That's because you are not indenting the comment, indentation is very important in `.sass`

Note how github renders the following examples with the "indented" `.sass` syntax:
```sass
.foo
// WRONG: bad indentation
  bar: baz
```
```sass
.foo
  // RIGHT: good indentation
  bar: baz
```

pamelalozano16: Gotcha, thanks!

source/documentation/at-rules/function.md
```diff
@@ -18,30 +18,30 @@ Functions are called using the normal CSS function syntax.
 [`@return` at-rule]: #return
 
 {% codeExample 'functions' %}
-  @function pow($base, $exponent) {
-    $result: 1;
-    @for $_ from 1 through $exponent {
-      $result: $result * $base;
+  @function fibbonacci($n) {
```
Goodwine, line 21: replace `fibbonacci` with `fibonacci`

source/documentation/at-rules/function.md
```diff
@@ -18,30 +18,30 @@ Functions are called using the normal CSS function syntax.
...
+      $sequence: append($sequence, $new);
     }
-    @return $result;
+    @return nth($sequence, length($sequence));
   }
 
   .sidebar {
     float: left;
-    margin-left: pow(4, 3) * 1px;
+    margin-left: fibbonacci(4)+px;
```
Goodwine, line 32: use `* 1px` instead of string concatenation `+px` to change units

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,123 @@
...
+introduction: >
+   Sass has added support for the CSS Values and Units Level 4 specification.
+   These global functions are now reserved: `round()`, `mod()`, `rem()`,
+   `sin()`, `cos()`, `tan()`, `asin()`, `acos()`, `atan()`, `atan2()`, `pow()`,
+   `sqrt()`, `hypot()`, `log()`, `exp()`, `abs()`, and `sign()`.
+-------------------------------------------------------------------------------
+{% compatibility 'dart: "1.65.0"', 'libsass: false', 'ruby: false' %}{% endcompatibility %}
+
+The new calc functions are now reserved as global functions. These function
+calls will be parsed CSS calculations instead of functions: [`round()`], [`mod()`],
```
Goodwine, line 12:
```suggestion
calls will be parsed as CSS calculations instead of function calls: [`round()`], [`mod()`],
```
(you may need to reflow to 80 chars)

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,123 @@
...
+.sidebar 
+    // WRONG, Sass parses `rem()` as a calculation, not as a call to the function
+    // declared above. This throws an error because `rem()` requires two parameters.
+    margin-left: rem(9) * 1px
+{% endcodeExample %}
+
+However, if the function is loaded from another file and called with a namespace
+it is parsed as a function call.
+
+{% codeExample 'calc-function-import' %}
```
Goodwine, line 59:
```suggestion
{% codeExample 'calc-function-use' %}
```

source/documentation/breaking-changes/calc-functions.md
```diff
@@ -0,0 +1,123 @@
...
+  }
+{% endcodeExample %}
+
+{% headsUp %}
+  Because calling a function from a module loaded [_without a namespace_] by
+  writing `@use <url> as *` has the same syntax as calling a global function,
+  Sass will parse these function calls as calculations, not as calls to the
+  function defined in the module!
+
+[_without a namespace_]: /documentation/at-rules/use/#choosing-a-namespace
```
Goodwine, line 97: Nit: match indentation
```suggestion
  [_without a namespace_]: /documentation/at-rules/use/#choosing-a-namespace
```

Goodwine (changes requested):

Goodwine (approved):

PR #822, opened 2023-09-05
fix(documentation): fix syntax.specialFunction typo

dancer1325:
* Remove an extra `#`, fixing the sample code

# How to review?
* [Running locally](https://github.com/sass/sass-site/blob/main/CONTRIBUTING.md#running-locally), checking that all is working fine
* Check in your preferred editor (online or not), that it's parsed to CSS
image


nex3:
This isn't a typo—the [`element()` function](https://developer.mozilla.org/en-US/docs/Web/CSS/element) takes as its input an ID selector, so the first `#` is necessary.

Issue #779, opened 2023-07-20
Playground improvements (click-to-copy, default initial code, movable panels, etc.)


jgerigmeyer:
Opening this issue to capture a number of proposed improvements to the new Sass Playground. Feel free to split this into separate issues as you see fit.

1. Default code contents on initial Playground load (if not following a direct link with existing contents)
2. Click-to-copy input SCSS/Sass or output CSS: see [design proposal](https://xd.adobe.com/view/84c872fe-0e51-46e4-a59c-f72aab2b8c6a-d19b/screen/413c2a93-be31-4b54-8bbe-ef30b8af2dd0/)
3. Link directly from code examples throughout to the site, to open the code in the Playground: see [design proposal](https://xd.adobe.com/view/84c872fe-0e51-46e4-a59c-f72aab2b8c6a-d19b/)
4. Drag to expand/collapse input/output/console panel widths: see [design proposal](https://xd.adobe.com/view/84c872fe-0e51-46e4-a59c-f72aab2b8c6a-d19b/screen/413c2a93-be31-4b54-8bbe-ef30b8af2dd0/)
5. Autocomplete Sass-specific keywords (e.g. `@use`, `@debug`) and in-scope variables
6. Click on line number in the Console to go to that line in the input panel
7. ~~Convert existing input between SCSS and Sass when switching between panels~~

nex3:
I'm striking through number 7 as "infeasible". We actually used to maintain a tool that could do this, but it turns out that the ongoing maintenance cost of needing to be able to produce nicely-formatted source code for every type of source AST node is substantial. I don't think it's worth the value of using it in the playground.

ravindUwU:
Thoughts on encoding the line number into the URL state and placing the cursor at that line when the playground loads? 🤔

nex3:
That's an interesting idea. I wonder if it would make more sense to copy the location specifically when there's _highlighted_ text—it seems like the most likely use case is to indicate "look at this part of the code", and highlighting might be a clearer signal that the user intends the position to be meaningful.

ravindUwU:
Oh that's way better!! I'd love to have a go at implementing this over the weekend 😄. I guess there are a few decisions to be made about 🤔,

1. The current URL state is `<inputFormat><outputFormat><content>` and doesn't include a content delimiter so I'm unsure if extending this format to include a selection is ideal... thoughts on making it a JSON object instead? Decoding can support both formats so that existing playground links won't break.
  
   ```typescript
[...]

PR #829, opened 2023-09-12
Add import math lib example code

bertyhell:
I never find how i need to import modules in scss anywhere in the docs

So propose we add it at the top of docs page

Maybe this can be added to other module pages as well in separate PR's

### DIFF
--- a/source/documentation/modules/math.md
+++ b/source/documentation/modules/math.md
@@ -4,6 +4,12 @@ title: sass:math
 
 {% render 'doc_snippets/built-in-module-status' %}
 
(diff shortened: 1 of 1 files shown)


nex3:
I think a better way to do this would be to add appropriate `@use` rules to each individual code example in the page, so that they're fully runnable in isolation. This also works well with the idea in https://github.com/sass/sass-site/issues/779 of linking examples directly to the playground, since otherwise these examples won't run in the playground. I've filed https://github.com/sass/sass-site/issues/830 to track that.

Issue #832, opened 2023-09-14
Wrong options are shown in the migrator documentation


stof:
On the website, the documentation at https://sass-lang.com/documentation/cli/migrator/#usage shows `-dry-run -verbose` as the options to use for dry runs. This is wrong as those options should be `--dry-run --verbose` (with 2 dashes for the beginning of long options).

However, the source of the documentation in this repo is right. So it looks like the build process is responsible for introducing this issue.

nex3:
It looks like the double hyphens are getting combined into U+2013 EN DASH. I don't think we get a lot of value out of Unicode ligature conversions—we should see if we can turn that off globally in the eleventy settings.

stof:
I managed to find where this comes from.

PR #835, opened 2023-09-15
Disable the typography replacements in the markdown engine

stof:
Closes https://github.com/sass/sass-site/issues/832

See https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/replacements.js for the list of replacements this feature is doing.
the documentation does not rely on any of them currently.


nex3 (approved):
Thanks!

PR #848, opened 2023-09-26
RFC for Package Importer proposal

jamesnw:


source/blog/039-rfc-package-importer.md
```diff
@@ -0,0 +1,118 @@
...
+
+would allow a package consumer to write:
+
+```scss
+@use "pkg:library";
+@use "pkg:library/utils";
+```
+
+The Node Package Importer also supports several other ways of defining entry
+points, which it resolves in the following order:
```
nex3, line 54: It's probably worth mentioning here that this covers the ways most Sass packages already define their entrypoints, so they can be loaded with `pkg:` without needing any changes. I know you mention that below, but it helps add context for why we support so many different mechanisms.

source/blog/039-rfc-package-importer.md
```diff
@@ -0,0 +1,118 @@
...
+
+[analysis]: https://github.com/oddbird/sass-pkg-test/tree/main/analysis
+[conditional exports]: https://nodejs.org/api/packages.html#conditional-exports
+[Vite]: https://github.com/vitejs/vite/pull/7817
+[Parcel]: https://github.com/parcel-bundler/parcel/blob/2d2400ded4615375ee6bd53ef77b4857ad1591dd/packages/transformers/sass/src/SassTransformer.js#L163
+[Sass Loader for Webpack]: https://github.com/webpack-contrib/sass-loader/blob/02df41203adfda96959e56abb43bd35a89ec11ba/src/utils.js#L514
+
+Users will need to opt-in to the new Package Importer by importing the
+`nodePackageImporter` from `sass` and including it in the list of `importers`.
+This won't be available for Sass in the browser.
```
nex3, line 109: Can you add a code example here?

nex3 (approved):
A couple small changes, but this looks good otherwise!

source/blog/039-rfc-package-importer.md
jamesnw: Added in https://github.com/sass/sass-site/pull/848/commits/b5310ddf28c1425732fd93bdf4ae1447971ec54e

source/blog/039-rfc-package-importer.md
jamesnw: Added in https://github.com/sass/sass-site/pull/848/commits/b5310ddf28c1425732fd93bdf4ae1447971ec54e

nex3 (approved):

PR #850, opened 2023-09-28
Reduce the size of the Black Lives Matter banner

nex3:
Most of the text here is outdated at this point. The Sass core team
wants to keep the banner itself both for its political purpose and
because it's been pragmatically effective at getting racists to refuse
to use Sass.

### DIFF
--- a/source/_data/alerts.yml
+++ b/source/_data/alerts.yml
@@ -1,4 +1,1 @@
-- heading: Black Lives Matter
-  body: >
-    Sass stands with the protesters against police violence. We encourage our
-    users to **get in the streets and join them if you can**.
+- body: '**Black Lives Matter**'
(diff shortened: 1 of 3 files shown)


Goodwine (approved):
I that it looks weird having the top and bottom black bars having different sizes, so I started playing with the UI and I think if you remove [`margin-bottom: var(--sl-gutter);`](https://github.com/sass/sass-site/blob/bcfd61ff92d7cf78aea95dd8cd75e0b8f784ea3d/source/assets/sass/components/_lists.scss#L24C24-L24C39), then it'll look better.

This is what I mean:
sass website screenshot showing layout misaligned due to margin-bottom


Deleting that value makes it look better, but you lose the bottom margin at the end of the page for the footer. I think that the margin should change so it only affects the footer.

sass website screenshot showing layout looks OK after removing margin-bottom

nex3:
I'd like to avoid making other design changes in this PR. It may be worth sending a separate PR for it, but it's probably a good idea to loop in @jina (the original designer) to get an expert opinion.

PR #945, opened 2024-02-07
docs: clarify @use's relation to selectors

Stan-Stani:
vs members

Maybe I wasn't reading with focus, but the current documentation did not make it clear to me that styles/selectors from non-.css files were loaded when using `@use`.


nex3:
I think the current wording ("combines CSS from multiple stylesheets") is more accurate. "Selectors" are just the part of a style rule that declares which element it matches, and style rules are far from the only thing that can be part of a stylesheet's CSS output.

Stan-Stani:
Some of my confusion comes from "combining CSS." Which made it sound like
@use could be used to load CSS (as in .css files) XOR to load members from
SASS.

It might just be me though.

PR #1003, opened 2024-03-27
Recommend sass-embedded gem

ntkme:
There are a lot of pain in Sass community caused by lack of wide adoption of dart-sass. Especially, for the maintainers of sass libraries (e.g. bootstrap), they have to deal with incompatibility between libsass and dart-sass or deprecation warnings, because as wildly used sass libraries they still need to support both libsass and dart-sass implementations in short term.

Today, the ruby sass-embedded gem already got a fair amount of adoption in ruby community. I think it's a good time to encourage it over sassc gem, which should help the community to move forward to an up-to-date implementation.

### DIFF
--- a/source/install.md
+++ b/source/install.md
@@ -33,7 +33,7 @@ Dart VM, so it's faster but only supports Windows, Mac OS, and Linux.
 
 There are also community-maintained wrappers for the following languages:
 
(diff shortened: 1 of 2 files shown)


nex3:
Oh, I hadn't realize this wasn't already recommended :sweat_smile:

nex3 (approved):

PR #1041, opened 2024-05-06
Add the css-function-mixin deprecation in the menu

stof:
#1028 has forgotten to update those when adding the new deprecation


nex3 (approved):

nex3:
Thanks!

PR #1060, opened 2024-05-21
Generate deprecations tables from Sass API

jathak:


source/documentation/cli/dart-sass.md
```diff
@@ -478,79 +478,7 @@ Remove this setting if you need to keep using this feature.
...
-    <td>Passing percentages to the Sass <code>abs()</code> function.</td>
-  </tr>
-  <tr>
-    <td><code>fs-importer-cwd</code></td>
-    <td>1.73.0</td>
-    <td>Using the current working directory as an implicit load path.</td>
-  </tr>
-</tbody>
-</table>
+{% deprecations 'active' %}{% enddeprecations %}
```
nex3, line 481: Should we list obsolete deprecations somewhere, so people at least know which names are valid?

nex3 (changes requested):

source/documentation/cli/dart-sass.md
jathak: We could, though there's not really a difference in behavior between passing an obsolete deprecation and an invalid one to any of the options (in both cases, you get a warning but no error). Do you think it's still worth listing them here?

nex3: I think it may still be helpful as a reference for users, yeah.

jathak: Okay, done (there's no change in the actual rendered site yet, but once there are obsolete deprecations, this will include the list of them)

nex3 (approved):

PR #1091, opened 2024-06-21
Include highlighted text in Playground URL state

### LINKED ISSUE #1090: Include highlighted text in Playground URL state
Forking this from https://github.com/sass/sass-site/issues/779#issuecomment-2171211045, and specifically responding to https://github.com/sass/sass-site/issues/779#issuecomment-2174754931:

> 1. The current URL state is `<inputFormat><outputFormat><content>` and doesn't include a content delimiter so I'm unsure if extending this format to include a selection is ideal... thoughts on making it a JSON array instead? Decoding can support both formats so that existing playground links won't break.
>    ```ts
>    type PlaygroundUrlState = [
>        /* content: */ string,
>        /* isScss: */ 0 | 1,
>        /* isExpanded: */ 0 | 1,
>        /* selection: */ [/* from: */ number, /* to: */ number] | undefined,
>    };
>    ```

Keeping the URL concise is a top priority, so I don't think JSON is the right way forward. I'd rather use something more like the standard means of adding metadata to URLs, like `https://sass-lang.com/playground/#<base64>?p=<line1>,<col1>:<line2>,<col2>`.

> 2. Should highlights be more visually distinctive? Maybe `#EFEB88`?
>    

I'm agnostic about this, although if we do choose a different color it should be one from among the [brand colors](https://sass-lang.com/styleguide/color/). @jina is the expert here.

ravindUwU:
Fixes #1090 😄🎉

Line numbers encoded as `L1C2-L3C4` instead of `1,2:3,4` because the former gets URI encoded verbatim, while the latter gets encoded as `1%2C2%3A3%2C4`.

### DIFF
--- a/source/assets/js/playground.ts
+++ b/source/assets/js/playground.ts
@@ -26,6 +26,7 @@ function setupPlayground() {
     compilerHasError: false,
     inputValue: hashState.inputValue || '',
     debugOutput: [],
+    selection: hashState.selection || null,
   };
 
   // Proxy intercepts setters and triggers side effects
@@ -41,7 +42,11 @@ function setupPlayground() {
       } else if (prop === 'inputValue') {
         debouncedUpdateCSS();
       }
-      if (['inputFormat', 'outputFormat', 'inputValue'].includes(prop)) {
+      if (
+        ['inputFormat', 'outputFormat', 'inputValue', 'selection'].includes(
+          prop
+        )
+      ) {
         debounceUpdateURL();
       }
       return set;
@@ -57,6 +62,10 @@ function setupPlayground() {
         if (v.docChanged) {
           playgroundState.inputValue = editor.state.doc.toString();
         }
+
+        if (v.selectionSet) {
+          playgroundState.selection = editorSelectionToStateSelection();
+        }
       }),
     ],
     parent: document.querySelector('.sl-code-is-source') || undefined,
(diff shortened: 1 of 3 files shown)


ravindUwU:
Hi @nex3 👋 can you have a quick look at this PR? 😅

The CI fails within `build-prod` → `build:typedoc` → `typedoc-build.sh` → [~:18](https://github.com/ravindUwU/sass-site/blob/main/tool/typedoc-build.sh#L18) `git checkout origin/main` and I don't quite understand why it fails here because it runs within `.language` which is a fresh clone of the Sass repo 🤔

```
1:32:23 PM: > sass-site@2.0.0 build:typedoc
1:32:23 PM: > ./tool/typedoc-build.sh
1:32:23 PM: + [[ ! -d .language ]]
1:32:23 PM: + cd .language
1:32:23 PM: + git fetch
1:32:23 PM: From https://github.com/sass/sass
1:32:23 PM:  * [new branch]      interleaved-decls -> origin/interleaved-decls
1:32:23 PM:    3b066ec..50c3ad5  main              -> origin/main
1:32:23 PM: + [[ -n '' ]]
1:32:23 PM: + git checkout origin/main
1:32:23 PM: error: Your local changes to the following files would be overwritten by checkout:
1:32:23 PM: 	package-lock.json
1:32:23 PM: Please commit your changes or stash them before you switch branches.
1:32:23 PM: Aborting
1:32:23 PM: ERROR: "build:typedoc" exited with 1.
1:32:24 PM: Failed during stage 'building site': Build script returned non-zero exit code: 2 (https://ntl.fyi/exit-code-2)
```

source/assets/js/playground.ts
```diff
@@ -68,11 +77,64 @@ function setupPlayground() {
...
+      fromLine.number,
+      range.from - fromLine.from + 1,
+      toLine.number,
+      range.to - toLine.from + 1,
+    ];
+  }
+
+  function restoreSelection() {
+    if (playgroundState.selection === null) {
+      return;
```
nex3: Should this clear the editor's selection if it has one?

nex3 (changes requested):

source/assets/js/playground.ts
ravindUwU: hmmm 🤔 The editor's always at L1C1 when it loads, but yes... I think clearing the selection here makes that function consistent the other `update*` functions. I'll also rename it to `updateSelection`. Thanks! 😊

nex3 (approved):
Thanks!

PR #1101, opened 2024-07-08
Document the mixed declaration breaking change

nex3:
See sass/sass#3846

### DIFF
--- a/source/documentation/breaking-changes/mixed-decls.md
+++ b/source/documentation/breaking-changes/mixed-decls.md
@@ -1,0 +1,118 @@
+---
+title: 'Breaking Change: Mixed Declarations'
+introduction: >
+  CSS is changing the way it handles declarations mixed with nested rules, and
+  we want to make sure Sass matches its behavior.
+---
+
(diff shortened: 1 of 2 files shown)


jathak (approved):

source/documentation/breaking-changes/mixed-decls.md
```diff
@@ -0,0 +1,118 @@
...
+    font-weight: normal;
+  }
+
+  .example a {
+    font-weight: bold;
+  }
+{% endcodeExample %}
+
+When [plain CSS Nesting] was first introduced, it behaved the same way. However,
+after some consideration, the CSS working group decided it made more sense to
```
yisibl, line 45: Could you add a CSSWG issue that would be helpful for authors to understand why the CSS syntax does what it does?

I think it should be: https://github.com/w3c/csswg-drafts/issues/8738

nex3: Good idea. I've sent out https://github.com/sass/sass-site/pull/1113

PR #1104, opened 2024-07-11
Add documentation for how to silence deprecations

nex3:


source/_includes/silencing_deprecations.liquid
```diff
@@ -0,0 +1,62 @@
...
+{% endheadsUp %}
+
+### Silencing Deprecations in Dependencies
+
+Sometimes, your dependencies have deprecation warnings that you can't do
+anything about. You can silence deprecation warnings from dependencies while
+still printing them for your app using the [`--quiet-deps` flag] on the command
+line, or the [`quietDeps` option] in the JavaScript API.
+
+[`--quiet-deps flag`]: /documentation/cli/dart-sass/#quiet-deps
```
jathak, line 40: Only `--quiet-deps` should be in ticks to fix this link (same with `--silence-deprecations` below)

nex3: Good catch!

PR #1111, opened 2024-07-16
Add click to copy input and output

jamesnw:
[Original design](https://xd.adobe.com/view/84c872fe-0e51-46e4-a59c-f72aab2b8c6a-d19b/screen/413c2a93-be31-4b54-8bbe-ef30b8af2dd0/)

image

Questions:
- [ ] Should we add click to copy for Debug console? This would likely not copy well, and is not in the initial design
- [ ] Should the buttons go to the left or right of the version and "CSS (Compiled)"? Those weren't present in original designs
- [ ] What should the copy notification be? Ideas:
  - Reuse existing copy message banner space, but show "Copy [URL/input/css] to clipboard"
  - Show copy message banners to the left of each button- would need to consider space & position


nex3:
Looks pretty good! Just a couple things I'd like to see changed:

* Can we make the copy icon color match the text color?
* Can we add a copy icon to the "Copy URL" button in the upper right, so that that icon is consistently associated with copying everywhere it appears?

jamesnw:
@nex3 Sorry about this! I opened this as a WIP in the wrong repo, before design had a chance with it. I'll make these changes, though!

PR #1121, opened 2024-07-24
Document Embedded Dart Sass

nex3:


source/dart-sass.md
```diff
@@ -104,5 +107,49 @@ as fast as the asynchronous API, due to the overhead of asynchronous callbacks.
 [brand new JavaScript API]: /documentation/js-api/
 [legacy API]: /documentation/js-api/#md:legacy-api
 
+## Embedded Dart Sass
+
+Dart Sass also supports the [Embeddded Sass protocol], which allows any
```
jathak, line 112:
```suggestion
Dart Sass also supports the [Sass embedded protocol], which allows any
```

(or possibly keep this with just the extra d dropped and change the name of the reference below?)

jathak (approved):

PR #1122, opened 2024-07-25
Natalie Weizenbaum

chaofix:
Removing "Free palestine" and firing Natalie Weizenbaum


dquak:
"No Politics or offense in public code." Litterly supporting terrorism in your main page

PR #1150, opened 2024-08-21
Update terror supporting yaml back to its origin

dimagoltsman:
Killing all jews in israel and calling for murder of israelies isnt the best way to achieve peace.
Being  a woke does not gives the right to yelling that all jews should be killed (like in the commit this was changed), also, ignoring black lives just to get the ability to be antisemitic is cruel.

### DIFF
--- a/source/_data/alerts.yml
+++ b/source/_data/alerts.yml
@@ -1,2 +1,2 @@
-- body: '**Free Palestine**'
+- body: '**Black Lives Matters**'
   color: black


borislit:
I totally agree. Free Palestine movement's slogan is "From the River to the Sea" which is a clear call for termination/transform of the jews. Not only is it cruel, it marginalizes people who have probably contributed to the success of sass.

MrBahur:
Palestine, or as it has been commonly known since 1948, Israel, is a democratic country with over 2 million Muslim Arabs living there. The phrase 'Free Palestine' is often associated with anti-Semitic and anti-Israel movements. I fully agree with the suggested changes in this PR and believe it should not have been approved in the first place

PR #1164, opened 2024-09-04
[Playground] Default code contents, Indented formatting

jamesnw:
Adds an initial Sass document, showing modules, lists, variables, and nesting. We also considered using the [variables example](https://sass-lang.com/guide/#variables), and can switch to that if preferred.

This also adds formatting when editing in indented mode.

Addresses # 1 from #779. 

[Scss](https://deploy-preview-1164--sass-lang.netlify.app/playground/)
[Indented](https://deploy-preview-1164--sass-lang.netlify.app/playground/#eJwzMAQAAJMAYg==)

If you have the default code in the editor, switching to the other syntax will load the default code in that syntax.


source/assets/js/playground/editor-setup.ts
```diff
@@ -34,6 +34,24 @@ import {
 } from '@codemirror/view';
 
 import {playgroundHighlightStyle} from './theme.js';
+import {EditorView} from 'codemirror';
+
+const syntax = new Compartment();
+
+const changeSyntax = (
```
nex3, line 41: Please document this, and declare functions using `function` syntax.

I should probably bring in our common TS static check suite here...

nex3 (approved):

source/assets/js/playground/editor-setup.ts
jamesnw: Updated, and I'm a fan of automating those checks!

nex3 (approved):

PR #1166, opened 2024-09-05
[Playground] Autocomplete

jamesnw:
Adds Sass-specific autocomplete to the [Playground](https://deploy-preview-1166--sass-lang.netlify.app/playground/), specifically-

- At-rules, like @use, @debug, etc. (CSS at rules will be added to the `@codemirror/lang-css` package in https://github.com/codemirror/lang-css/releases/tag/6.3.0 via #1172, not here)
- Built in module imports- `@use "sass:math"`, for instance
- Built in module namespaces only if module is imported- `math`
- Built in module variables only if module is imported- `math.$pi`
- Built in module functions only if module is imported- `math.clamp`


jamesnw:
@nex3 I have a few options for getting the current list of module functions and variables. Essentially, I'm able to generate a list of the values using Sass, `#{map.keys(meta.module-functions("math"))};`. To extract it from there, because this will be done in Node on build, I can't use DOM methods to get the values, which would have made this simpler. A few other options-

1. Use `@debug` to emit the value to a custom Logger (requires a Sass compilation (or two) per module, but simple parsing)
2. Use regex to parse the values from CSS. (simplest, but most prone to break)
3. Parse the CSS using `csstree` or `@csstools/css-parser-algorithms` which is already installed as a dependency of stylelint. (Most robust, but potentially overengineered)

Do you have any preference on this, or other ideas? I didn't see a clear way of parsing this from the existing documentation.

rollup.config.js
```diff
@@ -7,6 +8,11 @@ const {defineConfig} = require('rollup');
 
 const prod = process.env.BABEL_ENV === 'production';
 
+// Warn if module-members.ts has not been generated
+if (!existsSync('./source/assets/js/playground/module-members.ts')) {
+  throw new Error('module-members.ts is missing');
+}
+
```
jgerigmeyer, line 15: This shouldn't be required... but also doesn't hurt?

source/assets/js/playground/autocomplete.ts
```diff
@@ -6,7 +6,15 @@ import {
...
-import moduleMembers from './module-members';
+import type {ModuleDefinition} from './generate-module-members';
+
+let moduleMembers: ModuleDefinition[] = [];
+try {
+  moduleMembers = require('./module-members');
+} catch (error) {
+  console.error('module-members.ts is missing');
+  throw error;
+}
```
jgerigmeyer, line 17: @jamesnw Do you think this is worth it? It basically just allows the linter to run fine without needing the generated file. 🤷

nex3:
Can you pass the result of `map.keys()` to a custom function? That seems like the best way to get a JS data structure out of Sass.

jamesnw:
> Can you pass the result of `map.keys()` to a custom function? That seems like the best way to get a JS data structure out of Sass.

That worked well!

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+  name: string;
+  description?: string;
+};
+
+type CompletionModule = CompletionInfo & {
+  functions?: CompletionInfo[];
+  variables?: CompletionInfo[];
+};
+
+const moduleNames = moduleMembers.map(mod => mod.name);
```
nex3, line 65: This value reads weird—you're iterating over module _members_, but the variable name suggests that it's a _module_ rather than a member of one. I think this suggests that the name "module-members.ts" isn't really accurate. Maybe call it "module-metadata.ts" instead?

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+const moduleDescriptions = {
+  color:
+    'generates new colors based on existing ones, making it easy to build color themes',
+  list: 'lets you access and modify values in lists',
+  map: 'makes it possible to look up the value associated with a key in a map, and much more',
+  math: 'provides functions that operate on numbers',
+  meta: 'exposes the details of Sass’s inner workings',
+  selector: 'provides access to Sass’s powerful selector engine',
+  string: 'makes it easy to combine, search, or split apart strings',
+};
```
nex3, line 76: I'd move this into `generate-module-members.ts` and inject it into the generated module metadata. Then not only do you have a single chunk of data for each module, but you also only have one source of truth where you're manually listing all the known modules.

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+
+const builtinModules: CompletionModule[] = moduleMembers.map(modMember => {
+  let variables;
+  if ('variables' in modMember) {
```
nex3, line 80: Why not just let this be empty in modules with no variables?

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+    ':', // No end quote, on semicolon- `"sass:`
+    'PseudoClassName', // No end quote, after semicolon- `"sass:m`
+  ];
+  if (!potentialTypes.includes(nodeBefore.type.name)) return null;
+  const potentialParentTypes = [
+    'UseStatement', // When wrapped in quotes
+    'PseudoClassSelector', // No end quote, after the colon
+  ];
+  if (!potentialParentTypes.includes(nodeBefore.parent?.type.name || ''))
+    return null;
```
nex3, line 120: Style nit: use brackets when an `if` is multiple lines.

source/assets/js/playground/generate-module-members.ts
```diff
@@ -0,0 +1,92 @@
...
+    f: extract(map.keys(meta.module-functions($module)), $module, functions);
+    $variables: map.keys(meta.module-variables($module));
+    @if (list.length($variables) > 0){
+      v: extract(map.keys(meta.module-variables($module)), $module, variables);
+    }
+  }
+}
+`;
+
+  const modMap: ModuleDefinition[] = [];
```
nex3, line 46: Would it be cleaner to make this a `Record<string, ModuleDefinition>` with the module names as keys? Then instead of some `Array.prototype.reduce` stuff, you could use a simpler `mapRecord<I, O>(record: Record<T, I>, mapper: (I) => O): Record<T, O>` utility function.

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+    },
+    {}
+  )
+);
+
+const moduleFunctionsCompletions = Object.freeze(
+  builtinModules.reduce(
+    (acc: {[k: string]: CompletionResult['options'] | []}, mod) => {
+      acc[mod.name] =
+        mod.functions?.map(variable => ({
```
nex3, line 153: Don't all modules have a defined functions array?

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+          validFor: identifier,
+        })) || [];
+      return acc;
+    },
+    {}
+  )
+);
+
+// Returns the list of built in modules that are included in the text.
+function includedBuiltinModules(state: EditorState) {
```
nex3, line 168: Since the editor does seem to maintain some degree of a syntax tree, is it possible to do this in a slightly more robust way? It's not a big deal to have this have false positives, but if it's easy to iterate through top-level AST nodes looking for `@use` that might be better.

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+function builtinModuleItemCompletion(
+  context: CompletionContext
+): CompletionResult | null {
+  const nodeBefore = syntaxTree(context.state).resolveInner(context.pos, -1);
+  if (
+    ![nodeBefore.type.name, nodeBefore.parent?.type.name].includes(
+      'NamespacedValue'
+    )
+  )
+    return null;
```
nex3, line 209: Brackets

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+    mod => moduleFunctionsCompletions[mod]
+  );
+  const includedModVariables = includedModules.flatMap(
+    mod => moduleVariableCompletions[mod]
+  );
+
+  return {
+    from: moduleNameMatch.from,
+    to: moduleNameMatch.to,
+    options: [...includedModVariables, ...includedModFunctions],
```
nex3, line 228: Won't this suggest all members for _all_ included modules, not just the one under the current namespace?

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+    return null;
+
+  const moduleMatch = context.matchBefore(/["'](sass:)?\w*/);
+
+  if (!moduleMatch) return null;
+  if (moduleMatch.from === moduleMatch.to && !context.explicit) return null;
+  return {
+    from: moduleMatch.from + 1,
+    to: moduleMatch.to,
+    options: moduleCompletions,
```
nex3, line 129: Since we already have a function for determining which modules are loaded, it might be worth excluding already-loaded modules from these completions.

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,241 @@
...
+    validFor: identifier,
+  };
+}
+
+const moduleVariableCompletions = Object.freeze(
```
nex3, line 134: Probably for a follow-up PR, but it's worth thinking about ways to extract the documentation for variables and functions from `source/documentation/modules/*.md` so we can include them in these autocompletions.

source/assets/js/playground/autocomplete.ts
jamesnw: I switched to this in 55932c5feebbba79b253de2fa37cdfb5966d77a2.

source/assets/js/playground/autocomplete.ts
jamesnw: Added in https://github.com/sass/sass-site/tree/55932c5feebbba79b253de2fa37cdfb5966d77a2

source/assets/js/playground/generate-module-members.ts
jamesnw: Potentially it could be cleaner in the cases where I'm reducing things into `Record<ModuleName, CompletionSource>`, but the metadata is being used in other ways as well. I'm generally a fan of keeping data in flat arrays of objects, as it tends to be the most versatile.

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+    options: atRuleOptions,
+    validFor: identifier,
+  };
+}
+
+// A list of all the Sass built in modules.
+const moduleNames = moduleMetadata.map(mod => mod.name);
+type ModuleName = (typeof moduleNames)[number];
+
+// Matches any of the built in Sass modules.
```
nex3, line 64:
```suggestion
// Matches an identifier namespaced within any of the built in Sass modules.
```

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+  )
+);
+
+// Type predicate for modules names.
+function isModuleName(string?: string | null): string is ModuleName {
+  return moduleNames.includes(string as ModuleName);
+}
+
+// Returns the list of built in modules that are included in the text.
+function includedmoduleMetadata(state: EditorState): ModuleName[] {
```
nex3, line 158:
```suggestion
function includedModuleMetadata(state: EditorState): ModuleName[] {
```

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+    from: moduleMatch.from + 1,
+    to: moduleMatch.to,
+    options: notIncludedModuleCompletions,
+    validFor: identifier,
+  };
+}
+
+// Completion results for module variables.
+const moduleVariableCompletions = Object.freeze(
+  moduleMetadata.reduce(
```
nex3, line 121: If we're going to keep the modules as a list of objects, I'd make this clearer by creating a helper function like `mapModulesByName<V>(value: (module: moduleMetadata[number]) => V): Record<K, V>`

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+    (acc: {[k: string]: CompletionResult['options'] | []}, mod) => {
+      acc[mod.name] =
+        mod.variables?.map(variable => ({
```
nex3, line 124: If `variables` is now just an empty list for modules without them, you don't need the `?` here

source/assets/js/playground/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+      'NamespacedValue'
+    )
+  ) {
+    return null;
+  }
+  const moduleNameMatch = context.matchBefore(moduleNameRegExp);
+
+  if (!moduleNameMatch) return null;
+  if (moduleNameMatch.from === moduleNameMatch.to && !context.explicit)
+    return null;
```
nex3, line 217: Brackets

tool/generate-module-metadata.ts
```diff
@@ -0,0 +1,119 @@
+// Gets list of all functions and variables for the built-in modules
+//
+// Outputs `sass-site/source/assets/js/playground/module-metadata.ts` file, which is omitted from source code and
+// must be built before the `playground` bundle is built with rollup.
+import {compileString, sassTrue} from 'sass';
+import {writeFileSync} from 'fs';
+import path from 'path';
+
+const moduleDescriptions = {
```
nex3, line 9: Document top-level members

tool/generate-module-metadata.ts
```diff
@@ -0,0 +1,119 @@
...
+type ModuleName = (typeof modules)[number];
+
+interface ModuleDefinition {
+  name: ModuleName;
+  description: string;
+  functions?: string[];
+  variables?: string[];
+}
+
+// Generate Scss with a custom function that extracts each module's functions
```
nex3, line 39: Style nit:
```suggestion
// Generate SCSS with a custom function that extracts each module's functions
```

tool/generate-module-metadata.ts
```diff
@@ -0,0 +1,119 @@
...
+  'string',
+] as const;
+
+type ModuleName = (typeof modules)[number];
+
+interface ModuleDefinition {
+  name: ModuleName;
+  description: string;
+  functions?: string[];
+  variables?: string[];
```
nex3, line 36: Modules without functions or variables should have empty lists, not undefined lists.

```suggestion
  functions: string[];
  variables: string[];
```

tool/generate-module-metadata.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Generate Sass
+  const moduleSass = `
+    ${modules.map(mod => `@use "sass:${mod}";`).join('\n')}
+    $modules: ${modules.join(',')};
+
+    @each $module in $modules {
+    $_: extract($module, (
+          functions: map.keys(meta.module-functions($module)),
+          variables: map.keys(meta.module-variables($module))
+        ));
```
nex3, line 51: Style nit:

```suggestion
      $_: extract($module, (
        functions: map.keys(meta.module-functions($module)),
        variables: map.keys(meta.module-variables($module))
      ));
```

tool/generate-module-metadata.ts
```diff
@@ -0,0 +1,119 @@
...
+          variables: [],
+        };
+
+        if (functionStrings.length > 0) {
+          moduleDefinition.functions = functionStrings;
+        }
+
+        if (variableStrings.length > 0) {
+          moduleDefinition.variables = variableStrings;
+        }
```
nex3, line 87: It would be much simpler to write:

```suggestion
        const moduleDefinition: ModuleDefinition = {
          name,
          description: moduleDescriptions[name],
          functions: functions.asList
            .toArray()
            .map(key => key.assertString().text),
          variables: variables.asList
            .toArray()
            .map(key => key.assertString().text),
        };
```

nex3 (changes requested):

source/assets/js/playground/autocomplete.ts
jamesnw: Added in https://github.com/sass/sass-site/pull/1166/commits/ee7ac67c8677fd8b6bae2e863ff3086f55264165.

tool/generate-module-metadata.ts
jamesnw: Added in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

tool/generate-module-metadata.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

tool/generate-module-metadata.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

tool/generate-module-metadata.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

source/assets/js/playground/autocomplete.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

source/assets/js/playground/autocomplete.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

source/assets/js/playground/autocomplete.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

source/assets/js/playground/autocomplete.ts
jamesnw: Fixed in https://github.com/sass/sass-site/pull/1166/commits/7c4ef88299d155c3d4db97b36d502220278162a6

source/assets/js/playground/autocomplete.ts
```diff
@@ -115,37 +115,41 @@ function moduleMetadataCompletion(
     validFor: identifier,
   };
 }
+/**
```
nex3, line 118: Style nit: newlines between top-level members

source/assets/js/playground/autocomplete.ts
```diff
@@ -115,37 +115,41 @@ function moduleMetadataCompletion(
...
-        })) || [];
-      return acc;
-    },
-    {}
+  mapModulesByName(
+    mod =>
+      mod.variables.map(variable => ({
+        label: `${mod.name}.${variable}`,
+        type: 'variable',
+      })) || []
```
nex3, line 138:
```suggestion
      }))
```

source/assets/js/playground/autocomplete.ts
```diff
@@ -115,37 +115,41 @@ function moduleMetadataCompletion(
...
-    {}
+  mapModulesByName(
+    mod =>
+      mod.functions.map(func => ({
+        label: `${mod.name}.${func}`,
+        apply: `${mod.name}.${func}(`,
+        type: 'method',
+        boost: 10,
+        validFor: identifier,
+      })) || []
```
nex3, line 152:
```suggestion
      }))
```

nex3 (approved):

PR #1167, opened 2024-09-05
[Playground] Code links

jamesnw:
Adds "Edit in Playground" link to each code example in the [documentation](https://deploy-preview-1167--sass-lang.netlify.app/guide). The link will reflect the selected syntax.

If there are multiple sections in the example, for instance in the [Modules](https://deploy-preview-1167--sass-lang.netlify.app/guide/#modules) section, no link is shown. Those examples require importing and a file system, which isn't supported. I'm open to other behavior here if preferred.


nex3:
The functionality looks great, but I'm not sure about the styles. A few considerations that occur to me:

* The link is so much darker than everything else in the page, it really stands out to me in a way that seems contrary to its purpose.
* Since the focus of the examples and the user editing them is the Sass code, it seems like the playground link should be more visually associated with the source block than the output.
* Although the playground doesn't have its own logo, it _does_ have the Sass logo on screen at all times, so using that to visually identify the link with the playground page may be more immediately recognizable than pure text.

All that makes me think that something like this:

image

or even potentially like this:

image

would be a better fit.

As prior art here, this is what it looks like in the TypeScript documentation:

image


...where the "Try" button only shows up when the code is being hovered. I'd be fine with just copying that design as well.

source/_includes/code_examples/tab.liquid
```diff
@@ -1,3 +1,3 @@
-<li class="ui-tabs-tab{% if syntax == 'css' %} css-tab{% endif %}{% if enabled %} ui-tabs-active{% endif %}">
+<li class="ui-tabs-tab{% if syntax == 'css' %} css-tab{% endif %}{% if enabled %} ui-tabs-active{% endif %}" data-syntax="{{ syntax }}">
```
nex3, line 1: What is this data attribute for?

source/helpers/components/codeExample.ts
```diff
@@ -327,3 +333,34 @@ function getCanSplit(
...
 }
+
+// Create links to editable example in the playground. Returns false for
+// multiple sections/files, which are not supported by the playground.
+function generatePlaygroundLinks({
+  scss,
+  sass,
+}: {
+  scss: string[];
+  sass: string[];
```
nex3, line 344: These should both be `| false` to match the current semantics, although honestly I think `null` would be a better fit. I'm surprised tsc isn't complaining about this.

nex3 (approved):
Style looks good, except it would be nice to add a bit more space between the Sass logo and "Playground".

source/_includes/code_examples/tab.liquid
jgerigmeyer: Outdated -- removed in https://github.com/sass/sass-site/pull/1167/commits/4612ff139916ba78965ed2732f76c67a721b261f

source/helpers/components/codeExample.ts
jgerigmeyer: This is the type def for the input args, not the return value. Changed to `null` in https://github.com/sass/sass-site/pull/1167/commits/4612ff139916ba78965ed2732f76c67a721b261f

stacyk:
> Style looks good, except it would be nice to add a bit more space between the Sass logo and "Playground".

@nex3 I made a few updates here:
- Added a small gap between the logo and Playground
- Fixed bottom alignment (in some cases the links weren't touching the bottom of the panel)
- The language tabs arrow pseudo elements were sitting under the panel after I added the relative position, so I bumped those up (to avoid the gray border overlapping the arrow)
Screenshot 2024-09-09 at 2 15 10 PM

source/helpers/components/codeExample.ts
nex3: Oh I see. In that case, this should have a return type :slightly_smiling_face:

nex3 (approved):

source/helpers/components/codeExample.ts
jamesnw: Added in 2e22b0dae45bf3c3c28b3512877c6ec2ad0b09ce.

PR #1170, opened 2024-09-09
First draft, color space announcement

mirisuzanne:
- This includes some background on color spaces. I don't know if that's necessary or not. We could consider providing a link to skip that section.
- There are a few places I reused bits of text or examples from the request for comments, or the proposal. Hopefully that's ok. 
- There are a few examples where I didn't have any easy way to get the result. Those should be marked with `result: ???`. Sometimes I used the ColorJs library to get a result, which might not be right (e.g. might have different levels of precision). 
- I don't think I covered every minor change (like changes to `color.invert()`) but I tried to cover the essentials. 
- [x] I still need to add a few images …


source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+## New CSS formats give us the choice
+
+Moving forward, there are two directions we could go with wide gamut colors:
+
+- Color formats that re-fit larger and larger gamuts into simple coordinates, stretching the colors to preserve clean, geometric boundaries.
+- Color formats that maintain their *perceptually uniform* spacing, without any regard for specific gamuts.
+
+On the one hand, clean boundaries allow us to easily stay inside the range of available colors. Without those boundaries, it would be easy to *accidentally* request colors that aren't even physically possible. On the other hand, we expect these colors to be *perceived* by *other humans* -- and we need to make things *look* consistent, with enough contrast to be readable.
+
+The [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/) defines a number of new CSS color formats. Some of them maintain geometric access to specific color spaces. The `hwb()` function has been around for several years now, and defines `sRGB` colors using `hue`, `whiteness`, and `blackness` channels.  It's an interesting format, and [I've written about it before](https://www.miriamsuzanne.com/2022/06/29/hwb-clamping/).
```
nex3, line 65: The transition to talking about `hwb()` here feels a bit abrupt. Maybe a bridge sentence like "You're probably already familiar with the `rgb()` and `hsl()` functions." would help?

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+In the same `color()` function, we can also access the 'device independent' (and gamut-less) `xyz` color spaces -- often used as an international baseline for converting between different color models. I won't get into [white points](https://www.w3.org/TR/css-color-4/#white-point) here, but we can specify `xyz-d65` (the default) explicitly, or use `xyz-d50` instead.
+
+Working out from `xyz`, we get a number of new *theoretically unbounded* color formats -- prioritizing *perceptually uniform* distribution over clean geometry. These are available in functions of their own, including `lab()` (`lightness`, `a`, and `b`) and `lch()` (`lightness`, `chroma`, and `hue`) along with the newer 'ok' versions of each -- `oklab()` and `oklch()`. If you want the full history of these formats, [Eric Portis has written a great explainer](https://ericportis.com/posts/2024/okay-color-spaces/).
```
nex3, line 71: Nit, just to avoid confusion with "working out" as a verb in itself

```suggestion
Working outwards from `xyz`, we get a number of new *theoretically unbounded* color formats -- prioritizing *perceptually uniform* distribution over clean geometry. These are available in functions of their own, including `lab()` (`lightness`, `a`, and `b`) and `lch()` (`lightness`, `chroma`, and `hue`) along with the newer 'ok' versions of each -- `oklab()` and `oklch()`. If you want the full history of these formats, [Eric Portis has written a great explainer](https://ericportis.com/posts/2024/okay-color-spaces/).
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+## TL;DR top priority new formats
+
+For the color experts, it's great to have all this flexibility. For the rest of us, there are a few stand-out formats:
+
+- `color(display-p3 …)` provides access to a wider gamut of colors, which are available on many modern displays, while maintaining a clear set of gamut boundaries.
+- `oklch(…)` is the most intuitive and perceptually uniform space to work in, a newer alternative to `hsl(…)` -- `chroma` is very similar to `saturation`. But there are no guard rails here, and it's easy to end up outside the gamuts that any screen can display, or even outside the realm of physical reality.
```
nex3, line 78: I feel like "outside physical reality" might be too scary here. As long as users stay within normal lightness and saturation ranges, these should all at least be _theoretically_ meaningful colors, right?

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+The Sass color functions use the same syntax as the CSS functions, which means that a given color can be represented in a variety of different spaces. For example, these are all the same color:
+
+```scss
+$named: MediumVioletRed;
+$hex: #C71585;
+$hsl: hsl(322.2 80.91% 43.14%);
+$oklch: oklch(55.34% 0.2217 349.7);
+$display-p3: color(display-p3 0.716 0.1763 0.5105);
+```
```
nex3, line 97: Make this a proper example, maybe with `@debug`s so that it'll do something visible in the playground:

```suggestion
{% codeExample 'color-fns', false %}
  @debug MediumVioletRed;
  @debug #C71585;
  @debug hsl(322.2 80.91% 43.14%);
  @debug oklch(55.34% 0.2217 349.7);
  @debug color(display-p3 0.716 0.1763 0.5105);
  ===
  @debug MediumVioletRed
  @debug #C71585
  @debug hsl(322.2 80.91% 43.14%)
  @debug oklch(55.34% 0.2217 349.7)
  @debug color(display-p3 0.716 0.1763 0.5105)
{% endcodeExample %}
```

Same below.

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Sass colors hold their space
+
+Historically, both CSS and Sass would treat the different color-spaces as *interchangeable*. When all the color formats describe the same color gamut using the same underlying model, you can provide a color using `hsl()` syntax, and the parser can eagerly convert it to `rgb()` without risking any data loss. That's no longer the case for modern color spaces.
+
+In general, any color defined in a given space will remain in that space, and be emitted in that space. The space is defined by the function used, either one of the named spaced passed to `color()`, or the function name (e.g. `lab` for colors defined using the `lab()` function).
+
+However, the `rgb`, `hsl`, and `hwb` spaces will be considered "legacy spaces", and will often get special handling for the sake of backwards compatibility. Legacy colors will continue to be emitted in the most backwards-compatible format available. This matches CSS’s own backwards-compatibility behavior. Colors defined using hex notation or CSS color names are also considered part of the legacy `rgb` color space.
```
nex3, line 105: Nit: present tense

```suggestion
However, the `rgb`, `hsl`, and `hwb` spaces are considered "legacy spaces", and often get special handling for the sake of backwards compatibility. Legacy colors are still emitted in the most backwards-compatible format available. This matches CSS’s own backwards-compatibility behavior. Colors defined using hex notation or CSS color names are also considered part of the legacy `rgb` color space.
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+// result: oklch(55.34% 0.2217 349.7)
+$converted: color.to-space($brand, 'oklch');
+
+// results: oklch, false
+$new-space: color.space($brand);
+$is-legacy: color.is-legacy($brand);
+```
+
+Once we can convert a color between spaces, we no longer consider those colors to be *equal*. But we can ask if they would render as 'the same' color, using the `color.same()` function:
```
nex3, line 131:
```suggestion
Once we convert a color between spaces, we no longer consider those colors to be *equal*. But we can ask if they would render as 'the same' color, using the `color.same()` function:
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+$brand: hsl(0 100% 25.1%);
+
+// result: 25.1%
+$hsl-lightness: color.channel($brand, "lightness");
+
+// result: 37.67%
+$oklch-lightness: color.channel($brand, "lightness", $space: oklch);
+```
+
+CSS has also introduced the concept of 'powerless' and 'missing' color channels. For example, an `hsl` color with `100%` lightness will *always be white*. In that case, we can consider both the `hue` and `saturation` channels to be powerless. Changing their value won't have any impact on the resulting color. Sass allows us to ask if a channel is powerless using the `color.is-powerless()` function:
```
nex3, line 158: Actually, according to the current spec, only `0%` saturation causes the hue to become powerless. Lightness _never_ causes powerlessness. (This is actually wrong in our implementation too :cold_sweat:.)

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+@use 'sass:color';
+$brand: hsl(none 100% 25.1%);
+
+// result: false
+$missing-lightness: color.is-missing($brand, "lightness");
+
+// result: true
+$missing-hue: color.is-missing($brand, "hue");
+```
+
```
nex3, line 183: It's probably worth adding a note here that other than `is-missing()` and places where the CSS spec explicitly cares about missing channels, we treat them as though they had value 0.

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+Note that the returned color is still returned in the original color space, even when the adjustment is performed in a different space. That way we can start to use more advanced color spaces like `oklch` where they are useful, without necessarily relying on browsers to support those formats.
+
+The existing `color.mix()` function will also maintain existing behavior *when both colors are in legacy color spaces*. Legacy mixing is always done in `rgb` space. We can opt into other mixing techniques using the new `$method` parameter, which is designed to match the CSS specification for describing [interpolation methods](https://www.w3.org/TR/css-color-4/#interpolation-space) – used in CSS gradients, filters, animations, and transitions as well as the new CSS `color-mix()` function.
+
+For legacy colors, the method is optional. But for non-legacy colors, a method is required. In most cases, the method can simply be a color space name. But when we're using a color space with 'polar hue' channel (such as `hsl`, `hwb`, `lch`, or `oklch`) we can also specify the *direction* we want to move around the color wheel: `shorter hue`, `longer hue`, `increasing hue`, or `decreasing hue`:
```
nex3, line 204:
```suggestion
For legacy colors, the method is optional. But for non-legacy colors, a method is required. In most cases, the method can simply be a color space name. But when we're using a color space with "polar hue" channel (such as `hsl`, `hwb`, `lch`, or `oklch`) we can also specify the *direction* we want to move around the color wheel: `shorter hue`, `longer hue`, `increasing hue`, or `decreasing hue`:
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+```scss
+@use 'sass:color';
+
+// result: #660099
+$legacy: color.mix(red, blue, 40%);
+
+// result: ???
```
nex3, line 212: `rgb(176.2950613593, -28.8924497904, 159.1757183525)`

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+$lab: color.mix(red, blue, 40%, lab);
```
nex3, line 213: I think we should encourage users to use a named parameter for `$method`.

```suggestion
$lab: color.mix(red, blue, 40%, $method: lab);
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+// result: ???
```
nex3, line 215: `rgb(-129.55249236, 149.0291922672, 77.9649510422)`

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+$oklch-longer: color.mix(red, blue, 40%, oklch longer hue);
+```
+
+In this case, the first color in the mix is considered the 'origin' color. Like the other functions above, we can use different spaces for mixing, but the result will always be returned in that origin color space.
```
nex3, line 219:
```suggestion
In this case, the first color in the mix is considered the "origin" color. Like the other functions above, we can use different spaces for mixing, but the result will always be returned in that origin color space.
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+
+```scss
+@use 'sass:color';
+$extra-pink: color(display-p3 0.951 0.457 0.7569);
+
+// result: true, for display-p3 gamut
+$in-p3: color.is-in-gamut($extra-pink);
+
+// result: false, for srgb gamut
+$in-srgb: color.is-in-gamut($extra-pink, srgb);
```
nex3, line 241: For consistency, I think it makes sense to always recommend `$space:` be passed by name.

```suggestion
$in-srgb: color.is-in-gamut($extra-pink, $space: srgb);
```

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,279 @@
...
+$extra-pink: color(display-p3 0.951 0.457 0.7569);
+
+// result: ???
+$clip-to-srgb: color.to-gamut($extra-pink, srgb, clip);
+
+// result: ???
+$clip-to-srgb: color.to-gamut($extra-pink, srgb, local-minde);
+```
+
+All legacy and RGB-style spaces represent bounded gamuts of color. Since mapping colors into gamut is a lossy process, it should generally be left to browsers or done with caution. For that reason, out-of-gamut channel values are maintained by Sass whenever possible, even when converting into gamut-bounded color spaces. The only exception is that `hsl` and `hwb` color spaces are not able to express out-of-gamut color, so converting colors into those spaces will gamut-map the colors as well.
```
nex3, line 257: The last sentence here is no longer true. We now preserve out-of-gamut colors _everywhere_.

nex3 (changes requested):
> * This includes some background on color spaces. I don't know if that's necessary or not. We could consider providing a link to skip that section.

I like the idea of providing a link like "Already know all about color spaces? Skip ahead to the new Sass features!"

> * There are a few places I reused bits of text or examples from the request for comments, or the proposal. Hopefully that's ok.

Totally okay.

> * There are a few examples where I didn't have any easy way to get the result. Those should be marked with `result: ???`. Sometimes I used the ColorJs library to get a result, which might not be right (e.g. might have different levels of precision).

Ran these and added results.

> * I don't think I covered every minor change (like changes to `color.invert()`) but I tried to cover the essentials.

Yeah, we've got comprehensive docs and changelogs for that.

nex3:
Retargeting this to merge with my color 4 updates PR

source/blog/042-wide-gamut-colors-in-sass.md
mirisuzanne: I don't know if that's guaranteed, because the range relationship `100% == 0.4` doesn't adjust based on hue/lightness, but the actual Chroma available is dependent on hue/lightness. Still, maybe 0.4 is low enough to be safe across the board? @jamesnw is looking into it.

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,285 @@
...
+
+Sass now accepts all the new CSS formats, and treats them as first-class *colors* that we can manipulate, mix, convert, and inspect. These functions are all available globally:
+
+- `lab()`, `oklab()`, `lch()`, and `oklch()`
+-  `color()` using the `sRGB`, `srgb-linear`, `display-p3`, `a98-rgb`, `prophoto-rgb`, `rec2020`, `xyz`, `xyz-d65`, and `xyz-d50` color spaces
+- `hwb()` (Sass previously had a `color.hwb()` function, which is now deprecated in favor of the global function)
+
+The Sass color functions use the same syntax as the CSS functions, which means that a given color can be represented in a variety of different spaces. For example, these are all the same color:
+
+{% codeExample 'color-fns', false %}
```
mirisuzanne, line 91: @nex3 Should I move all the code demos to this format?

mirisuzanne:
- [x] Merged in `sass:color-4`
- [x] Accepted all the review suggestions
- [x] Addressed all the specific comments
- [x] Included a skip-to-features link
- [ ] Still a few unknown function results in the final code sample
- [ ] Do we want all code samples in dual formats using the `codeExample` helper?

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+We can also use the `color.to-gamut()` function to explicitly move a color so that it is in a particular gamut. Since there are several options on the table, and no clear sense what default CSS will use long-term, this function currently requires an explicit `$method` parameter. The current options are `clip` (as is currently applied by browsers) or `local-minde` (as is currently specified):
+
+```scss
+@use 'sass:color';
+$extra-pink: color(display-p3 0.951 0.457 0.7569);
+
+// result: ???
+$clip-to-srgb: color.to-gamut($extra-pink, srgb, clip);
```
nex3, line 260: `color(display-p3 0.930165971 0.4548604319 0.7563239804)`

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -0,0 +1,288 @@
...
+
+// result: ???
+$map-to-srgb: color.to-gamut($extra-pink, srgb, local-minde);
```
nex3, line 263: `color(display-p3 0.930165971 0.4548604319 0.7563239804)`, same as above :/

source/blog/042-wide-gamut-colors-in-sass.md
nex3: Yep!

source/blog/042-wide-gamut-colors-in-sass.md
```diff
@@ -118,109 +120,182 @@ Sass provides a variety of tools for inspecting and working with these color spa
...
+  $brand: hsl(0 100% 25.1%)
+
+  // result: 25.1%
+  @debug color.channel($brand, "lightness")
+
+  // result: 37.67%
+  @debug color.channel($brand, "lightness", $space: oklch)
+{% endcodeExample %}
+
+CSS has also introduced the concept of 'powerless' and 'missing' color channels. For example, an `hsl` color with `0%` saturation will *always be grayscale*. In that case, we can consider both the `hue` and `saturation` channels to be powerless. Changing their value won't have any impact on the resulting color. Sass allows us to ask if a channel is powerless using the `color.is-powerless()` function:
```
nex3, line 199:
```suggestion
CSS has also introduced the concept of 'powerless' and 'missing' color channels. For example, an `hsl` color with `0%` saturation will *always be grayscale*. In that case, we can consider the `hue` channel to be powerless. Changing its value won't have any impact on the resulting color. Sass allows us to ask if a channel is powerless using the `color.is-powerless()` function:
```

nex3 (approved):

PR #1177, opened 2024-09-10
[Playground] Click to Copy

jamesnw:
Adds Click to Copy to input and output panes in the playground.

Addresses item 2 in #779.


source/_includes/playground/header.liquid
nex3, line 1: I really like making the existing copy button match the new visual copy indicators :smiley:

source/assets/js/playground.ts
```diff
@@ -193,15 +194,53 @@ function setupPlayground(): void {
     const copyURLButton = document.getElementById('playground-copy-url');
     const copiedAlert = document.getElementById('playground-copied-alert');
 
-    let timer: undefined | number;
+    type Timer = undefined | number;
```
nex3, line 197: Please put type definitions at the top level and document them.

source/assets/js/playground.ts
```diff
@@ -193,15 +194,53 @@ function setupPlayground(): void {
...
+    };
+
+    function showCopiedAlert(msg: string): void {
+      if (!copiedAlert) return;
+      copiedAlert.innerText = msg;
+      copiedAlert.classList.add('show');
+      if (alertTimer) clearTimeout(alertTimer);
+      alertTimer = window.setTimeout(() => {
+        copiedAlert.classList.remove('show');
+      }, 3000);
```
nex3, line 213: This this number is used in multiple places now, it's probably a good idea to make it a top-level constant.

source/assets/sass/config/color/_content.scss
```diff
@@ -31,3 +31,7 @@ $sl-color--info-lighter: color.adjust(brand.$sl-color--patina, $lightness: 47%);
 $sl-color--error: #cf0254;
 $sl-color--warn: #c14e00;
 $sl-color--success: #168073;
+
+// Darker Shades of existing colors for use on a light gray background
+$sl-color--text-medium-dark: #5e6470; // darker version of pale-sky
+$sl-color--action-dark: #a12d69; // darker version of action
```
nex3, line 37: Should these be defined in terms of `color.scale()` or `color.adjust()`?

nex3 (changes requested):

source/assets/js/playground.ts
jamesnw: Resolved in https://github.com/sass/sass-site/pull/1177/commits/661dfef9188f7dccc54ad4b78209526cefadb94d

source/assets/js/playground.ts
jamesnw: Resolved in https://github.com/sass/sass-site/pull/1177/commits/661dfef9188f7dccc54ad4b78209526cefadb94d

source/assets/sass/config/color/_content.scss
stacyk: @nex3 I can update that. I just quickly used a color app's down arrows until it passed. I can switch to a more readable format and look up the exact adjustments to use to make it work based on the brand colors.

nex3: I'd appreciate it! We may also move to new color spaces for color conversions at some point, at which point knowing the underlying logic is valuable because we can just make the results more accurate.

jamesnw: Added in https://github.com/sass/sass-site/pull/1177/commits/93a4f1a6c70e1daa194be2ba97c98374716ccdf9

nex3 (approved):

PR #1178, opened 2024-09-12
Fix footer not showing the latest versions.

### LINKED ISSUE #1176: Broken latest version links for Current Releases in footer
It appears a recent deployment to the Sass site has broken the latest version links that used to appear next to each of the libraries in the Current Releases section in the site footer. Looking at recent commits, it looks like [4fad520e](https://github.com/sass/sass-site/commit/4fad520e645e090523ee346f8175dbf7c99b2c5b) is the most likely culprit as functions relating to release version information were reworked in it.

Goodwine:
Eleventy expects that default exports are provided with 'module.exports ='. Not with 'export default'.

Fixes #1176

### DIFF
--- a/source/_data/releases.ts
+++ b/source/_data/releases.ts
@@ -106,7 +106,7 @@ async function getLatestVersion(repo: string): Promise<string> {
 /**
  * Returns the version and URL for the latest release of all implementations.
  */
-export default async function (): Promise<Record<string, Release>> {
+module.exports = async function (): Promise<Record<string, Release>> {
   const repos = ['sass/libsass', 'sass/dart-sass', 'sass/migrator'];
   const cache = await getCacheFile();
 
@@ -129,4 +129,4 @@ export default async function (): Promise<Record<string, Release>> {
   }
 
   return data;
-}
+};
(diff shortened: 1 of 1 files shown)


source/_data/releases.ts
```diff
@@ -106,7 +106,7 @@ async function getLatestVersion(repo: string): Promise<string> {
 /**
  * Returns the version and URL for the latest release of all implementations.
  */
-export default async function (): Promise<Record<string, Release>> {
+module.exports = async function (): Promise<Record<string, Release>> {
```
nex3, line 109: Can you add a comment here indicating why `export default` doesn't work? It's what we use everywhere else, so the difference here is conspicuous.

Goodwine: Done, I wasn't sure if this is an issue with Eleventy or LiquidJS, but I documented it as an issue with Liquid

Goodwine: I ended up saying it's Eleventy because AFAICT LiquidJS itself doesn't support JS/TS files as `_data`, that's an Eleventy's feature

nex3 (approved):

PR #1179, opened 2024-09-12
[Playground] Debug links

jamesnw:
## Description
In the debug console, line numbers are now links to select or position the cursor at the relevant issue. Also, deprecation messages with links are now clickable.

## Related Issue(s)
Addresses number 6 in #779.

## Steps to test/reproduce

1. [Debug, Warn, and Deprecation links](https://deploy-preview-1179--sass-lang.netlify.app/playground/#eJwzNHRISU0qTVfQMDQqqNA3KajQtOYCAEPTBfY=?s=L1C9-L1C17)
2.[ Error link](https://deploy-preview-1179--sass-lang.netlify.app/playground/#eJwzNFRJz89PsVJISkyx5uJKrOZSUEjOz8kvslJQKUoFCtUCAKdZCcA=?s=L4C10-L4C14)
3. [Error without a range](https://deploy-preview-1179--sass-lang.netlify.app/playground/#eJwzNExJTUkqTVcAAA4tAu4=)
4. [Long content, more examples](https://deploy-preview-1179--sass-lang.netlify.app/playground/#eJztls1qhDAUhfd5isuMLQr+JU4pKIXprq8RNaNhUg2JM4yUefdGHYvSdRfFbEK4OSfnfjebYHy8aAY7TbVOBdfdLkOLStGKVpkSOpYsv1TgYiJv0UHePFObi6MopELW1N0zXL6WxMsgigCvFa2kBe96V1W5S3DsAyEvw4J9iMODN3nMbnZNNz5cbySeBL+qphHn1DZdoDtanFP4YOLKOl5QH94VpyJDjlT8k6o+GPtIYZ8kiek/b8sevhDAyj4MIaRSsqZ0Fwc+aNroQDPFTwYeYDhKl9YM3RGi44WPnHWuSQR4Tuv2ytQg+pFN09EFFcxdW3xwBK/qrmHmLYDET2PyfQiyyBbZIltki2yRLbJFtsgW+U+R51/5qgGEyCpfsTLb6Hz+dfMbfC+LvAXkLTJb5C0gfwM06Ea3)

Links should allow you to open in new tab/window with CMD/Shift/Right click, etc.


source/assets/js/playground.ts
```diff
@@ -155,6 +155,12 @@ function setupPlayground(): void {
     }
   }
 
+  function goToSelection(selection: PlaygroundState['selection']): void {
```
nex3, line 158: Please document all functions.

source/assets/js/playground.ts
```diff
@@ -258,8 +264,21 @@ function setupPlayground(): void {
       '.sl-c-playground__console'
     ) as HTMLDivElement;
     console.innerHTML = playgroundState.debugOutput
-      .map(displayForConsoleLog)
+      .map(item => displayForConsoleLog(item, playgroundState))
       .join('\n');
+    console.querySelectorAll('a.console-location').forEach(link => {
+      (link as HTMLAnchorElement).addEventListener('click', event => {
+        if (!(event.metaKey || event.altKey || event.shiftKey))
+          event.preventDefault();
```
nex3, line 272: Use brackets for multi-line `if`s

source/assets/js/playground.ts
nex3, line 158: The type `PlaygroundState['selection']` is re-used often enough it's probably worth making its own named type.

source/assets/js/playground/console-utils.ts
```diff
@@ -45,36 +59,92 @@ function lineNumberFormatter(number?: number): string {
...
+export function displayForConsoleLog(
+  item: ConsoleLog,
+  playgroundState: PlaygroundState
+): string {
+  const data: {
+    type: string;
+    lineNumber?: number;
+    message: string;
+    safeLink?: string;
+    range?: PlaygroundState['selection'];
```
nex3, line 79: Unless there's a meaningful distinction between this being `null` and `undefined`, it should probably be
```suggestion
    range: PlaygroundState['selection'];
```

source/assets/js/playground/console-utils.ts
```diff
@@ -45,36 +59,92 @@ function lineNumberFormatter(number?: number): string {
...
+): string {
+  if (!range) return '';
+  return serializeState({...playgroundState, selection: range});
+}
+
+export function displayForConsoleLog(
+  item: ConsoleLog,
+  playgroundState: PlaygroundState
+): string {
+  const data: {
```
nex3, line 74: Why store this as an object rather than distinct variables?

source/assets/js/playground/console-utils.ts
```diff
@@ -45,36 +59,92 @@ function lineNumberFormatter(number?: number): string {
...
+          parseInt(match[2]),
+          parseInt(match[1]),
+          parseInt(match[2]),
+        ];
+      }
     }
     data.lineNumber = lineNumber;
+
+    if (item.type === 'warn' && item.options.deprecationType?.id) {
+      data.safeLink = `https://sass-lang.com/d/${item.options.deprecationType.id}`;
```
nex3, line 134: I find the indirection in this safeLink thing a bit confusing, and I worry it'll get worse if we end up wanting to add more HTML to the messages. Consider instead having separate `htmlMessage` and `message` variables and injecting `htmlMessage ?? encodeHTML(message)`.

nex3 (changes requested):

source/assets/js/playground.ts
jamesnw: Completed in https://github.com/sass/sass-site/pull/1179/commits/2d7ba875d1154d13cd07cdf950058ec55826d158

source/assets/js/playground.ts
jamesnw: Added in https://github.com/sass/sass-site/pull/1179/commits/2d7ba875d1154d13cd07cdf950058ec55826d158

source/assets/js/playground.ts
jamesnw: Done in https://github.com/sass/sass-site/pull/1179/commits/2d7ba875d1154d13cd07cdf950058ec55826d158

source/assets/js/playground/console-utils.ts
jamesnw: Done in https://github.com/sass/sass-site/pull/1179/commits/2d7ba875d1154d13cd07cdf950058ec55826d158

source/assets/js/playground/console-utils.ts
jamesnw: I found it slightly easier to track initially, but it doesn't have a useful purpose now, removed in https://github.com/sass/sass-site/pull/1179/commits/2d7ba875d1154d13cd07cdf950058ec55826d158

source/assets/js/playground/console-utils.ts
jamesnw: This wouldn't work in this situation, unless I misunderstand. I'll explain what I'm doing here a bit more to see if there's a clearer and more future-proof method.

1. `message` comes from the Logger, and is inherently unsafe
2. When we have a deprecationType, we can generate a known safe URL for more info
3. If that known safe URL exists as a string in `message`, we want to turn that string into a link. 

I realized the link augmentation is a bit confusing to include in the htmlEncode function, so I moved it out in  https://github.com/sass/sass-site/pull/1179/commits/e86f9b9fe5a5926f8fcceb7930dbf43886a67400.

source/assets/js/playground.ts
```diff
@@ -155,7 +154,8 @@ function setupPlayground(): void {
     }
   }
 
-  function goToSelection(selection: PlaygroundState['selection']): void {
+  /** Updates state, applies selection in editor, and focus editor. */
```
nex3, line 157: It's fine for documentation comments to be sentence fragments that starts with verbs, but otherwise please use proper grammar. If you refer to parameters or other code objects as proper nouns, use `{@link ...}` to refer to them. Also, try to focus on what the semantics of the function are rather than just describing its implementation in prose. For example: `Highlights {@link selection} in the code editor.`

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
         span.end.column + 1,
       ];
     }
-    data.message = item.error?.toString() || '';
+    message = item.error?.toString() || '';
   } else if (['debug', 'warn'].includes(item.type)) {
-    data.message = item.message;
-    let lineNumber = item.options.span?.start?.line;
+    message = item.message;
+    let _lineNumber = item.options.span?.start?.line;
```
nex3, line 88: I'd like to clean up this logic and get rid of the confusing `lineNumber`/`_lineNumber` distinction.

`SourceSpan.start` is always defined, so `_lineNumber` will by undefined if and only if `item.options.span` is undefined. That means that instead of `if (typeof _lineNumber === 'undefined')` you could just make that an `else` clause on `if (item.options.span)`. At which point you can just assign `lineNumber` directly rather than using an intermediate variable for it.

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
         span.end.column + 1,
       ];
     }
 
-    if (typeof lineNumber === 'undefined') {
+    if (typeof _lineNumber === 'undefined') {
       const stack = 'stack' in item.options ? item.options.stack : '';
       const needleFromStackRegex = /^- (\d+):(\d+) /;
       const match = stack?.match(needleFromStackRegex);
       if (match?.[1]) {
```
nex3, line 104: It's odd to have two separate `if` clauses here, given that if the regular expression matches at all both of its groups will definitely be defined.

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
         span.end.line + 1,
         span.end.column + 1,
       ];
     }
 
-    if (typeof lineNumber === 'undefined') {
+    if (typeof _lineNumber === 'undefined') {
       const stack = 'stack' in item.options ? item.options.stack : '';
       const needleFromStackRegex = /^- (\d+):(\d+) /;
       const match = stack?.match(needleFromStackRegex);
```
nex3, line 103:
```suggestion
      const match = item.options.stack?.match(/^- (\d+):(\d+) /);
```

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
-  } = {
-    type: item.type,
-    lineNumber: undefined,
-    message: '',
-    safeLink: undefined,
-    range: null,
-  };
+  const type = item.type;
+  let lineNumber = undefined;
+  let message = '';
```
nex3, line 70:
```suggestion
  let message: string;
```

If you avoid giving this a default value, the compiler will ensure that it's never accessed before it's given a real value.

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
+  let safeLink = undefined;
```
nex3, line 71: Keeping around a `safeLink` variable still has most of the brittleness of the previous structure. What if we just replace `message` with `html` to begin with, HTML-encode it on assignment, and then inject the link directly instead of assigning `safeLink`?

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
+      lineNumber = span.start.line;
+      range = [
         span.start.line + 1,
         span.start.column + 1,
         span.end.line + 1,
         span.end.column + 1,
       ];
     }
-    data.message = item.error?.toString() || '';
+    message = item.error?.toString() || '';
```
nex3, line 85: Always use `??` rather than `||` for null-safety, to avoid weird truthiness interactions.
```suggestion
    message = item.error?.toString() ?? '';
```

source/assets/js/playground/console-utils.ts
```diff
@@ -36,115 +36,112 @@ export type ConsoleLog = ConsoleLogDebug | ConsoleLogWarning | ConsoleLogError;
...
+
   return `<div class="console-line">${locationStart}<span class="console-type console-type-${
-    data.type
-  }">@${data.type}</span>:${lineNumberFormatter(
-    data.lineNumber
-  )}${locationEnd}<div class="console-message">${encodeHTML(data.message, data.safeLink)}</div></div>`;
+    type
+  }">@${type}</span>:${lineNumberFormatter(
+    lineNumber
+  )}${locationEnd}<div class="console-message">${html}</div></div>`;
```
nex3, line 146: Does this end up with a dangling `:` if `lineNumber` is undefined?

nex3 (changes requested):

source/assets/js/playground/console-utils.ts
jamesnw: It does, only if a warning does not have a span, or does not include a parseable line number in its stack. I wasn't sure how to replicate that to verify, but I fixed that in b738fd44297af0931d22eca1bb3981607097b4e0.

source/assets/js/playground.ts
jamesnw: Updated in ee7c1d5e11c255736b0c1fda68bfcfcb8d5f8672

nex3 (approved):

PR #1187, opened 2024-09-18
Sort color functions alphabetically

FloEdelmann:
`color.channel` was sorted after `color.complement` rather than after `color.change`.


nex3 (approved):

nex3:
Thanks!

PR #1188, opened 2024-09-18
Fix unprefixed `color.adjust()` name

FloEdelmann:
With namespace, it's `color.adjust`, but without, it's `adjust-color`, not `adjust`.


nex3:
Rather than making reference to the soon-to-be-deprecated global functions, these should refer to `color.adjust()`.

nex3 (approved):

PR #1190, opened 2024-09-19
Remove nonexisting `color.saturate()` function

FloEdelmann:
Only the non-prefixed (deprecated) `saturate()` function exists, as described a few lines below in the "Heads up" block.


nex3:
Thanks!

PR #1204, opened 2024-09-27
[Playground] Resize panels

jamesnw:
This addresses number 4 in #779.

It adds a dependency on [`@spectrum-web-components/split-view`](https://opensource.adobe.com/spectrum-web-components/components/split-view/). While this web component was the best fit we found as far as lightweight and fully featured, it did limit the design, so this does not match our original designs. We're happy to talk about design options or tradeoffs as needed.

https://deploy-preview-1204--sass-lang.netlify.app/playground/


nex3:
I'm not intrinsically opposed to this design, but it does seem like a lot of change to the current style for a relatively limited use-case. Would it be possible to do something with the `resize` property that's a little less usable for people who want to do resizing but looks better in the common case where people don't?

stacyk:
Updates include:
- White gap between panels (as opposed to the previous thick gray line with slightly larger gripper in the middle)
- Focus background color matches link hover color

We could add a hover color if we want to, or keep it white.

source/assets/js/playground/split-view.ts
```diff
@@ -0,0 +1,17 @@
+import {SplitView} from '@spectrum-web-components/split-view';
+import '@spectrum-web-components/split-view/sp-split-view.js';
+
+/**
+ * Make split views responsive and stack vertically on narrow screens.
+ */
+export default function setupSplitView(): void {
```
nex3, line 7: Nit:

```suggestion
export default function setUpSplitView(): void {
```

"setup" is the noun form, "set up" is the verb :slightly_smiling_face:

nex3 (approved):
Oh, if the split-view component supports this look then I'm totally happy with the design :slightly_smiling_face:

source/assets/js/playground/split-view.ts
jamesnw: Updated!

nex3 (approved):

PR #1213, opened 2024-10-03
[Playground] use CSS resize for sizing playground panels

stacyk:
Testing the difference between CSS resize and the SplitView web component in  https://github.com/sass/sass-site/pull/1204

This addresses number 4 in https://github.com/sass/sass-site/issues/779

TODO: 
- [ ] Allow block resize on smaller screen sizes
- [ ] Set max block size for compiled panel
- [ ] Set better default inline sizes

https://deploy-preview-1213--sass-lang.netlify.app/playground/

### DIFF
--- a/source/assets/sass/components/_playground.scss
+++ b/source/assets/sass/components/_playground.scss
@@ -164,8 +164,8 @@ $playground-base-colors: (
   min-inline-size: 100%;
 
   @include config.sl-breakpoint--large {
-    min-inline-size: var(--sl-min-inline-size--playground-panel);
     max-inline-size: calc(100vw - var(--sl-min-inline-size--playground-panel));
+    min-inline-size: var(--sl-min-inline-size--playground-panel);
     resize: inline;
   }
 }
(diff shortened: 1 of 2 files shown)


stacyk:
@nex3 This isn't complete, but we wanted to get your feedback on this direction (compared to the SplitView web component in https://github.com/sass/sass-site/pull/1204). [You pointed out](https://github.com/sass/sass-site/pull/1204#issuecomment-2380255572) this might be a little less usable, ie. target area being less noticeable, quite a bit smaller, no way to use keyboard nav to control, etc.

We could also make edits to PR 1204 to have the same white gap between panels and make the gripper area white and the same size as the gutter. This would make it look much more similar to the current production site. We could add a hover/focus/active color if we want to, or keep it white. 

Let us know which direction you would prefer.

nex3:
Thanks for demoing this! I'm totally happy with https://github.com/sass/sass-site/pull/1204 now that the gutters are back to the original look. I don't think it needs a hover color.

PR #1260, opened 2024-11-26
RFC Blogpost for indented syntax

jamesnw:


source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,121 @@
...
+part of the language.
+
+## The new syntax
+
+Instead of adding syntax that is completely new to Sass, we propose to borrow
+syntax from the SCSS format and make it available in the indented syntax. This
+reduces the learning curve, code complexity, and likelihood of future conflicts
+with new CSS features.
+
+A common way of using this syntax will be wrapping lists with linebreaks in
```
jgerigmeyer, line 24: Nit: `line break` instead of `linebreak` (also other instances below)

```suggestion
A common way of using this syntax will be wrapping lists with line breaks in
```

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,121 @@
...
+  .item-#{ ²
+    $item
+  }
+    content: $item * ³
+        10
+```
+
+1. A statement can't end after the word "in" in an `@each` statement.
+2. A statement can't end inside the curly braces in an interpolation.
+3. A statement can't end after a multiplication operator.
```
jgerigmeyer, line 66: I think these references might be easier to parse if they used line numbers instead of superscript. Or I'm not sure if there's a better way to make it clear that they're not part of the code example?

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,121 @@
...
+
+// Back to indented syntax
+strong
+  font:
+    weight: bold
+```
+
+## Semicolons are also allowed
+
+We are also proposing that semicolons can optionally be added at the end of
```
jgerigmeyer, line 100: Maybe find another phrase rather than repeating "We are also proposing...". Maybe "Another proposed change would allow..." or something?

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,121 @@
...
+  display: grid;
+  grid-template: "header" min-content "main" 1fr;
+}
+```
+
+Powering this change is the underlying rule that linebreaks only end statements
+in places where statements can end. Anywhere where a statement can't end, a
+linebreak is treated as whitespace. This means that if the parser can't tell
+whether or not a linebreak is intended to end a statement, it will end the
+statement.
```
jgerigmeyer, line 50: Would it be helpful to also include a code example showing the "parser can't tell" path?

jgerigmeyer (commented):
👍

source/blog/044-rfc-indented-syntax.md
jamesnw: Line numbers made it harder to follow, I think.

I'll push a version with inline comments-
image

Here's the original-
image

source/blog/044-rfc-indented-syntax.md
jamesnw: Added in https://github.com/sass/sass-site/pull/1260/commits/2002663d4101d067f1190c39d0daa86ced8ab622

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+.grid {
+  display: grid;
+  grid-template: "header" min-content "main" 1fr;
+}
+```
+
+Powering this change is the underlying rule that linebreaks only end statements
+in places where statements can end. Anywhere where a statement can't end, a
+linebreak is treated as whitespace. This means that if the parser can't tell
+whether or not a linebreak is intended to end a statement, it will end the
```
nex3, line 49: I think asking the user to understand whether the parser can or can't tell something may be a bit confusing. Possibly better to say "if a linebreak _could_ end a statement, it _will_ end the statement".

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+statement.
+
+This example demonstrates several places where line breaks will be possible.
```
nex3, line 52: Even line breaks that end statements are "possible". Maybe better to say "will not end statements".

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+  .item-#{ 
+    $item /* A statement can't end inside the curly braces in an interpolation. */
+  }
+    content: $item * /* A statement can't end after a multiplication operator. */
+        10
+```
+
+
+As a counter example, here are some places in the same stylesheet where
+linebreaks could end statements, so linebreaks will cause compilation errors.
```
nex3, line 66: I think it's worth explaining a little more explicitly that these will cause compilation errors because the *next line* can't be parsed as a valid statement.

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+
+Another proposed change would allow semicolons to optionally be added at the end
+of statements in the indented syntax. Indentation rules still apply, so you
+won't be able to separate multiple statements on the same line, even with a
+semicolon (unless you're in a [SCSS-in-Sass](#scss-in-sass) block).
+
+## No breaking changes
+
+These changes are opt-in, so authors who don't want to use the new syntax do not
+have to make any changes or learn new syntax. No changes need to be made to
```
nex3, line 117: It's a bit disingenuous to say that users who don't use this won't have to learn it; it's still fairly likely they'll encounter other people's stylesheets that do use it.

source/blog/044-rfc-indented-syntax.md
```diff
@@ -0,0 +1,130 @@
...
+existing stylesheets written in the Indented Syntax.
+
+## Next steps
+
+This is still in the proposal phase, and we are open to feedback. Review the
+proposal [on
+Github](https://github.com/sass/sass/blob/main/proposal/indented-syntax-improvements.md),
+and [open an issue](https://github.com/sass/sass/issues/new) with any thoughts
```
nex3, line 125: Nit: use reference links

nex3 (changes requested):

nex3 (approved):

PR #1261, opened 2024-11-29
legacy-js-api page: Update Vite reference with Vite 6 having Modern API as default

joaopslins:
With Vite 6 recently released, I was looking at the docs and noticed that we can update this description.
I'm not sure about the language so please feel free to suggest copy improvements. Thanks!

Reference: https://vite.dev/guide/migration.html#sass-now-uses-modern-api-by-default


nex3 (approved):
Thanks!

PR #1290, opened 2024-12-31
[Playground] Add background to gutter

cr7pt0gr4ph7:
Make the background of the editor gutter non-transparent again, which turns this:

sass-lang com_playground_

into this (note the gutter on the left side of the editors):

sass-lang com_playground_ (1)


nex3:
Thanks!

Issue #1333, opened 2025-03-07
Document support for the new attr() function syntax


nex3:
See https://github.com/sass/sass/issues/4030

- [ ] Deprecation
- [ ] Final changes

PR #1340, opened 2025-03-11
Document support/deprecations for new `attr()` function syntax

nex3:
See #1333


source/documentation/values/strings.md
Goodwine, line 1: Are the changes on this file intended to go along with the breaking change announcement?

It seems related, but it doesn't mention `type()` 🤔

Goodwine (approved):

source/documentation/values/strings.md
nex3: Yes, this is all a single change as described in [the attr-type proposal](https://github.com/sass/sass/blob/main/accepted/attr-type.md).

PR #1358, opened 2025-04-17
Update meta.md to include `mixins` as a return type of `meta.type-of`

janis-me:
A very tiny change, but during development I spent a bit too long to figure out that  the `meta.type-of` of `meta.get-mixin(myMixin)` is not `function` but rather `mixin`. Makes a lot of sense, of course, but I didn't know it existed.

This PR makes the docs a bit more complete, I think.

Edit: I know you added the note `New possible values may be added in the future` in the docs. So if you feel that this list should not be extended explicitly, feel free to just close this :)


nex3:
Thanks!

nex3:
Oh, if you wouldn't mind opening a PR, I'd appreciate it.

PR #1445, opened 2025-08-22
Work around some Color.js bugs that are breaking the playground

### LINKED ISSUE #1444: Certain colors break the playground
It seems that certain color values break the ability for the Sass playground to properly take input and execute the Sass code. To reproduce, open a new playground session, delete the default text, and paste in `@debug rgb(10, 20, 30, 0.1)`. Note that the cursor immediately disappears and isn't possible to move accurately with the keyboard, and various other edits will fail in various strange ways. I believe this is because of errors somewhere in the `Color.js` library which are in turn not being captured correctly.

nex3:
Closes #1444

### DIFF
--- a/source/assets/js/playground/color-decorator.ts
+++ b/source/assets/js/playground/color-decorator.ts
@@ -89,7 +89,9 @@ class ColorWidget extends WidgetType {
       // Make sure errors like #1444 don't completely break interactivity.
       console.error(error);
       const errorMessage =
-        typeof error === 'object' && 'message' in error ? error.message : error;
+        typeof error === 'object' && error && 'message' in error
+          ? error.message
+          : error;
       return colorSwatchView('transparent', {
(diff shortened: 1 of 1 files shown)


source/assets/js/playground/color-decorator.ts
```diff
@@ -77,14 +82,27 @@ class ColorWidget extends WidgetType {
   }
 
   toDOM(): HTMLDivElement {
-    return colorSwatchView(this.color.toString(), Boolean(this.inP3));
+    let colorString: string;
+    try {
+      colorString = this.color.toString();
+    } catch (error) {
+      // Make sure errors like #1444 don't completely break interactivity.
+      console.error(error);
```
Goodwine, line 90: Do we want to log the error tho?

Goodwine (approved):

Issue #1446, opened 2025-08-23
Document misplaced rest arguments


nex3:
See https://github.com/sass/sass/issues/4120

- [x] Deprecation (https://github.com/sass/sass-site/pull/1447)
- [ ] Removal

PR #1447, opened 2025-08-23
Document misplaced rest arguments

nex3:
See https://github.com/sass/sass/issues/4120
See https://github.com/sass/sass-site/issues/1446

### DIFF
--- a/source/documentation/breaking-changes/misplaced-rest.md
+++ b/source/documentation/breaking-changes/misplaced-rest.md
@@ -1,0 +1,54 @@
+---
+title: 'Breaking Change: Misplaced Rest Arguments'
+introduction: >
+  Sass has historically allowed rest arguments to appear anywhere in an
+  argument list, even though they're always evaluated at the end. This closes
+  that loophole and requires rest arguments to appear at the end of argument
+  lists.
+---
+
+{% compatibility 'dart: "1.91.0"', 'libsass: false', 'ruby: false' %}{% endcompatibility %}
+
+[Rest arguments] in Sass (written `$args...`) were always intended to be written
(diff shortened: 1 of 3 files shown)


source/documentation/breaking-changes/misplaced-rest.md
```diff
@@ -0,0 +1,54 @@
...
+[Rest arguments] in Sass (written `$args...`) were always intended to be written
+at the end of argument lists. The way they're evaluated reflects this: they're
+always run after all the other arguments, and they're always added to the end of
+the list of positional arguments.
+
+[Rest arguments]: https://sass-lang.com/documentation/at-rules/function/#taking-arbitrary-arguments
+
+However, due to an oversight in the implementation of both Ruby Sass and Dart
+Sass, this was never enforced. It was possible to write a function call like
+`rgb([1, 2]..., 3)` with the rest argument before a positional (or named)
```
Goodwine, line 21: Just out of curiosity I tried passing 2 lists with `...` and I do get an error, but it's not very descriptive.
```
Plain CSS functions don't support keyword arguments.
  ╷
4 │   b: g([1, 2]..., [3, 4]...)
```

Maybe this is fine, but I just wanted to let you know. This is definitely outside of the scope of the current fix anyways

Goodwine (approved):

source/documentation/breaking-changes/misplaced-rest.md
stof: @Goodwine the error in your code snippet is because you used `g()` instead of the name of a Sass function. The error you get is unrelated to the current deprecation (it is related to the restriction of calls for CSS functions).

PR #1593, opened 2026-05-08
Add smoke test for playground

Goodwine:
Adds puppeteer to run the devserver and see if the playground loads correctly using a preselected input via a predefined URL.


tool/test-playground.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+    await new Promise(resolve => setTimeout(resolve, options.pollRate));
+  } while (Date.now() - startTime < options.timeout);
+  throw new Error('Timeout waiting for server to start.');
+}
+
+/** Starts the dev sever in the background. */
+async function startSever(): Promise<ChildProcess> {
+  console.log('Starting dev server...');
+  return spawn('npm', ['run', 'serve'], {stdio: 'ignore', detached: false});
```
Goodwine, line 42: I added `detached:false` because calling `kill()` seemed to leave some processes running in the background locally..

Having said that, that didn't help lol

tool/test-playground.ts
```diff
@@ -0,0 +1,113 @@
...
+async function main(): Promise<void> {
+  let serverProcess: ChildProcess | undefined;
+  try {
+    serverProcess = await startSever();
+    await waitForServer(BASE_URL);
+    await checkPlayground(`${BASE_URL}/${PLAYGROUND_PATH}`);
+  } finally {
+    if (serverProcess && !serverProcess?.killed) {
+      console.log('Shutting down dev server...');
+      serverProcess.kill('SIGKILL');
```
Goodwine, line 106: I tried SIGKILL and SIGINT, but that didn't help terminate the devserver locally 🤔 

I also tried adding a "sleep" for 3 seconds and seeing if it was killed, which it was, but I still got `npm run watch:...` instances running 🤔

tool/test-playground.ts
```diff
@@ -0,0 +1,114 @@
...
+/** Starts the dev sever in the background. */
+async function startSever(): Promise<ChildProcess> {
+  console.log('Starting dev server...');
+  return spawn('npm', ['run', 'serve'], {stdio: 'ignore', detached: false});
+}
+
+/**
+ * Waits for an element matching `selector` to contain `text`.
+ *
+ * Throws an error if such element is not found. Supports all options of
```
nex3, line 48:
```suggestion
 * Throws an error if such an element is not found. Supports all options of
```

tool/test-playground.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (!foundText)
+    throw new Error(
+      `Could not find an element "${selector}" containing "${text}"`,
+    );
+}
+
+/**
+ * Loads the playground and makes sure it loads correctly.
+ *
+ * We just thest that the playground loads the default code and that it has
```
nex3, line 76:
```suggestion
 * We just test that the playground loads the default code and that it has
```

nex3 (commented):
Would [killing the process group](https://stackoverflow.com/questions/392022/whats-the-best-way-to-send-a-signal-to-all-members-of-a-process-group) work?

It looks like the current build failure is due to the browser itself not starting, though. Surely someone has got puppeteer working on GitHub Actions before...

PR #1604, opened 2026-05-23
Migrate from bufbuild/buf-setup-action to bufbuild/buf-action

ntkme:
`buf-setup-action` has been deprecated: https://github.com/bufbuild/buf-action#migrating-from-individual-buf-actions


jathak (approved):
Thanks!

PR #1668, opened 2026-09-16
[meta.load] Add documentation for first-class modules and other APIs

nex3:
See https://github.com/sass/sass/issues/739


source/documentation/modules/meta.md
```diff
@@ -23,93 +23,60 @@ title: sass:meta
...
-      @if $border-contrast
-        border-color: #dadbdf
     ---
     // style.sass
     @use "sass:meta"
 
     body.dark
-      $configuration: ("border-contrast": true)
-      @include meta.load-css("dark-theme/code", $with: $configuration)
+      @include meta.load-css("dark-theme/code")
```
jathak: Is this supposed to also be `@include meta.css(meta.load("dark-theme/code")` to match the above?

jathak (approved):

source/documentation/modules/meta.md
nex3: Yes, good catch

